PR #1, opened 2017-11-02
Fix Oinkmaster modifysid with group name - v1.

jasonish:
Make the group name matcher the last matcher to be parsed,
and accept the match spec if it ends with .rules and not
prefixed with group.

This was broken while fixing up other issues in group name
parsing to make it more predictable.


victorjulien:
Tests ok.

jasonish:
Merged.

PR #2, opened 2017-11-02
rule testing: don't fail because of ASAN leak warnings

victorjulien:
Describe changes:
- don't fail if binary is ASAN enabled and reports memleak errors


jasonish:
Merged. Thanks!

PR #6, opened 2017-11-03
github/codeowners: add OISF/core-team

victorjulien:
Add core team so all PRs get a reviewer assigned. If the PR is by @jasonish it will be just OISF/core-team, otherwise it'll be both. Then @jasonish can approve in name of core-team as well.


jasonish:
Merged. Thanks!

PR #12, opened 2017-12-01
source index support - v2

jasonish:
Adds support for a source index. The index is an online file containing known sources that can easily be enabled. To support working with sources the following commands have been added:
- update-sources - downloads the index
- enable-source - enable a source from the index
- disable-source - disable an enabled source
- remove-source - remove a source, just removes the configuration; the source stays in the index
- add-source - add a custom source with a URL

Embedded support for ET/Pro has been dropped. Its now in the index. Support for ET/Open still exists as the default ruleset if no other is specified.

The "sources" field in /etc/suricata/update.yaml is now just a list of URLs, as we'll use the index for named sources.

Quick start guide is here: https://github.com/jasonish/suricata-update/wiki/Quick-Start

Previous PR: https://github.com/OISF/suricata-update/pull/11

Changes from last PR:
- Quick Start guide is now in the doc/ directory.
- Custom user agent.


jasonish:
Merged with a small fix for Python 3.

PR #13, opened 2017-12-04
better default source handling + doc - v1

jasonish:
The main feature of this PR is better default source handling.

When no sources are configured, the default remains to use et/open.

When enabling the first source, and /var/lib/suricata/update/sources is writable, enable the default source (currently defined as et/open). This is skipped if the source being enabled is et/open, or replaces et/open (for example, et/pro - https://github.com/jasonish/suricata-intel-index/blob/master/index.yaml#L25).

Also, when enabling a source, remove any sources that it replaces. So when et/pro is enabled, et/open will be removed.

In addition to the above:
- All sub-commands have been moved into their own source file.
- Add documention for all sub-commands, even if a minimal stub for now.


jasonish:
See https://github.com/OISF/suricata-update/pull/14

PR #15, opened 2017-12-05
version 1.0.0a1 - v1

jasonish:
Sets the version to 1.0.0a1. Updates doc bits since we'll push to PyPI.

Also include the sample update.yaml in the distribution package, forgot to do this earlier.


jasonish:
Closing. See https://github.com/OISF/suricata-update/pull/16

PR #17, opened 2017-12-08
Various fixes - v1

jasonish:
Issue https://redmine.openinfosecfoundation.org/issues/2334.
- Based on user feedback. --output would change the cache location, not just the rule output location. To fix this introduce -D (--data-dir) to change the /var/lib/suricata location and make the cache default to /var/lib/suricata/update/cache. This way the rule output can be redirected independent of the suricata-update working directory.
- Switching the config to a module as a singleton Python idiom helped make this easier to avoid passing a config object around everywhere. This is a Pythonic thing to do.

Issue https://redmine.openinfosecfoundation.org/issues/2336.
- Automatically run update-sources when list-sources is run and no index exists.

Also add more integration type tests with the main focus of testing user scenarios with more versions of Python.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@ _work
 
 # The file containing the git revision.
 /suricata/update/revision.py*
+
+# Working files for the tox testing framework.
+/.tox
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,4 @@ python:
   - "3.5"
   - "3.6"
 
-script: nosetests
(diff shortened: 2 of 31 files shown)


jasonish:
Closing. See https://github.com/OISF/suricata-update/pull/18

jasonish:
> Tests OK. Wonder if for the few existing users the move of the index should be handled more gracefully.

I thought about that. Just don't want alpha release mistakes leaving on forever. Suppose its code that could be removed post 1.0.

PR #34, opened 2018-03-25
Ubuntu trusty 1

jasonish:
Fix for Ubuntu Trusty which uses Python < 2.7.9 and doesn't have ssl.create_default_context, and by default doesn't verify certificates.

The fix is to attempt to use the newer ssl.create_default_context where we can control certificate validation, but fallback to the default.


pevma:
Works as expected on Trusty.

PR #36, opened 2018-04-30
bundled with suricata fixes - v1

jasonish:
Attempt to get default directories from the defaults.py module that suricata will drop when compiled (with Python support). In order for suricata-update to find the defaults.py, it must be installed as bundled with Suricata, or installed into the same prefix. The main idea here is to better support the use case when bundled.

Plus some other minor fixups.


victorjulien:
Tests OK standalone, but I guess I need to test the bundling as well, right?

PR #44, opened 2018-10-03
Use assertEqual instead of deprecated assertEquals

shivan1b:

Describe changes:
Replace the use of `assertEquals` in all places with `assertEqual` as the former was deprecated in https://bugs.python.org/issue9424
`assertEqual` goes well with Python 2.7 alongwith Python 3.x so this does not break any existing test cases.


jasonish:
Merged. Thanks!

PR #45, opened 2018-10-03
Add Python 3.7 to tox configurations

shivan1b:
- [x] I have read the contributing guide lines at
  https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [x] I have signed the Open Information Security Foundation
  contribution agreement at
  https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the
  changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: None

Describe changes:
Current tox configuration was missing Python 3.7 because of which tests were never run for a Python 3.7 environment. Fix that by adding `py37` to the configurations.

PR #46, opened 2018-10-04
Add Python 3.7 to tox configurations

shivan1b:

Describe changes:
Current tox configuration was missing Python 3.7 because of which tests were
never run for a Python 3.7 environment. Fix that by adding py37 to the configurations.

This PR is a revision of #45 having the requested changes.


jasonish:
Merged. Thanks.

PR #51, opened 2018-10-23
update classification.conf to Suricata 4.1 style - v1

jasonish:
This is a sync from Travis to make the test classification.config match the current style used in Suricata.

Oops, should not have done this on the OISF repo, but leaving as-is since already done.


jasonish:
Merged. Thanks Travis!

PR #52, opened 2018-10-24
Shorten the state of being written for rule files

shivan1b:
In an attempt to have shorter window for rule files of being written,
write the rules at first to `/tmp` and then copy the file to the actual
rule location.

Closes Redmine ticket #2327

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2327

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -592,7 +592,7 @@ def build_report(prev_rulemap, rulemap):
     return report
 
 def write_merged(filename, rulemap):
-
+    tmp_file = tempfile.NamedTemporaryFile(prefix="suricata-update_")
     if not args.quiet:
         prev_rulemap = {}
         if os.path.exists(filename):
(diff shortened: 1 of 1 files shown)


shivan1b:
@jasonish I'm not quite sure if the use of `shutil.copy` serves the actual purpose. I'm still to look up how is copy working underneath.
Making use of move did not make sense either with `tempfile`.
Do you have any suggestions?

shivan1b:
Ran a few tests, copy operation takes about 4 times lesser time than writing to the file.

suricata/update/main.py
```diff
@@ -608,10 +608,13 @@ def write_merged(filename, rulemap):
...
                         len(report["modified"])))
-    
-    with io.open(filename, encoding="utf-8", mode="w") as fileobj:
+
+    with io.open(tmp_file.name, encoding="utf-8", mode="w") as fileobj:
         for rule in rulemap:
             print(rulemap[rule].format(), file=fileobj)
+    # Copying the rule file from /tmp to the actual destination
+    shutil.copy(tmp_file.name, filename)
+    tmp_file.close()
```
jasonish: After thinking about this a bit, I think we can do without the temporary directory and copy. For this case we are outputing a single file, to a directory that we can assume we have write access to, and that file name is always suricata.rules as we hard code it for now. So we could in:
- use "suricata.rules.tmp" as a the temp file, in the same directory as the final file.
- then os.rename suricata.rules.tmp to suricata.rules
which should be faster, and safer.

Something similar could probably be done for the write_to_directory. But I'm not exactly sure how that would look at the moment. Needs a bit more thought I think.

shivan1b: About `write_to_directory`, since those files also do have a definite name, we could use the same scheme, I think.

PR #53, opened 2018-10-30
Add summary for updated sources

shivan1b:
`suricata-update` when used with its option `update-sources` did not
give any information about what changed and in which source. Add the
logs for any change in the sources, namely, addition, removal or change
in the content.
The log output now looks like:
```
$ ./bin/suricata-update update-sources
31/10/2018 -- 00:03:52 - <Info> -- Loading /etc/suricata/update.yaml
31/10/2018 -- 00:03:52 - <Info> -- Using data-directory /var/lib/suricata.
31/10/2018 -- 00:03:52 - <Info> -- Using Suricata configuration /etc/suricata/suricata.yaml
31/10/2018 -- 00:03:52 - <Info> -- Using /etc/suricata/rules for Suricata provided rules.
31/10/2018 -- 00:03:52 - <Info> -- Found Suricata version 4.1.0-dev at /usr/sbin/suricata.
31/10/2018 -- 00:03:52 - <Info> -- Downloading https://www.openinfosecfoundation.org/rules/index.yaml
31/10/2018 -- 00:03:53 - <Info> -- Source et/open was added
31/10/2018 -- 00:03:53 - <Info> -- Source empty/something was removed
31/10/2018 -- 00:03:53 - <Info> -- Source et/pro was changed
31/10/2018 -- 00:03:53 - <Info> -- Saved /var/lib/suricata/update/cache/index.yaml
```
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2472

### DIFF
--- a/suricata/update/commands/updatesources.py
+++ b/suricata/update/commands/updatesources.py
@@ -19,6 +19,7 @@ from __future__ import print_function
 import os
 import logging
 import io
+import yaml
 
 from suricata.update import config
 from suricata.update import sources
@@ -27,11 +28,36 @@ from suricata.update import exceptions
 
 logger = logging.getLogger()
 
+
 def register(parser):
     parser.set_defaults(func=update_sources)
 
+
+def compare_sources(initial_content, final_content):
+    initial_sources = initial_content.get("sources")
+    final_sources = final_content.get("sources")
+    added_sources = {source: final_sources[source]
+                     for source in final_sources if source not in initial_sources}
+    removed_sources = {source: initial_sources[source]
+                       for source in initial_sources if source not in final_sources}
+    if initial_content == final_content:
+        logger.info("No change in sources")
+        return
+    if added_sources:
+        for source in added_sources:
+            logger.info("Source %s was added", source)
+    if removed_sources:
+        for source in removed_sources:
+            logger.info("Source %s was removed", source)
+    for source in set(initial_sources) & set(final_sources):
+        if initial_sources[source] != final_sources[source]:
+            logger.info("Source %s was changed", source)
+
+
 def update_sources():
     local_index_filename = sources.get_index_filename()
+    with open(local_index_filename) as stream:
+        initial_content = yaml.safe_load(stream)
     with io.BytesIO() as fileobj:
         url = sources.get_source_index_url()
         logger.info("Downloading %s", url)
@@ -49,4 +75,7 @@ def update_sources():
                 return 1
         with open(local_index_filename, "wb") as outobj:
             outobj.write(fileobj.getvalue())
+        with open(local_index_filename) as stream:
+            final_content = yaml.safe_load(stream)
+        compare_sources(initial_content, final_content)
         logger.info("Saved %s", local_index_filename)


suricata/update/commands/updatesources.py
```diff
@@ -27,11 +28,36 @@
...
 
+
 def register(parser):
     parser.set_defaults(func=update_sources)
 
+
+def compare_sources(initial_content, final_content):
+    initial_sources = initial_content.get("sources")
+    final_sources = final_content.get("sources")
+    added_sources = {source: final_sources[source]
```
CuriousLearner: Just a minor improvement here:

How about we take the extra items beforehand like:

```removed_sources = dict(set(initial_sources.items()) - set(final_sources.items()))```

and

```added_sources = dict(set(final_resources.items()) - set(initial_sources.items()))```

IMHO, it would be more readable and pythonic.

File: suricata/update/commands/updatesources.py [hunk 2]
```diff
@@ -27,11 +28,36 @@
...
+                     for source in final_sources if source not in initial_sources}
+    removed_sources = {source: initial_sources[source]
+                       for source in initial_sources if source not in final_sources}
+    if initial_content == final_content:
```
CuriousLearner: Also, we can move line 43-45 on line 37.

This is initial check with the arguments itself on the equality. This way, we can avoid unnecessary computation, in case the dict of `initial_content` and `final_content` is long enough.

Thoughts?

suricata/update/commands/updatesources.py
shivan1b: Yes, this looks so much better. Thanks  a lot!

suricata/update/commands/updatesources.py
shivan1b: I do not think it is going to be very long anytime but doesn't hurt to avoid extra computations. Taking this into account too.

CuriousLearner:
@shivan1b You don't need to close PR's to address the comments. You can just push to the same branch and PR would be updated.

This way it would be easier for anyone to keep track of changes made during the review process.

shivan1b:
@CuriousLearner if there has to be any smallest of change, we need to create a new PR. These are the contribution rules for this organization. https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing

I'll be referencing this PR in the new one and resolving the discussion here afterward.

suricata/update/commands/updatesources.py
shivan1b: Turns out this does not work well for nested dictionaries. From all things I can find, I would end up processing more. Do you have some suggestion that works well for nested dictionaries for Python 2 as well as Python 3?

CuriousLearner: Hey, on giving a second thought to my suggestion, I realized it won't have the same result that you are expecting here, because set difference right now will compare both key and value in the tuples. In order to get the same result you would have to also add something like:

```extra_items = set(initial_sources.keys()) - set(final_resources.keys())
{k:v for k, v in initial_sources.items() if k in extra_items}```

which isn't making much sense anymore, given the already written solution.

I think just iterating like `{k: v for k, v in whatever_source}` will make it clear that the dict is just being filtered, or just using filter with lambda or just doing the following:
`{ k : second_dict[k] for k in set(second_dict) - set(first_dict) }`

PR #56, opened 2018-10-31
Fix --test-command; logging fixups - v1

jasonish:
Suricata-Update would fail on a custom --test-command when setting up the environment if no --output directory was explicitly passed. Fix this by using the proper way to get the output directory.

Based on feedback from @victorjulien, provide a better log message when failing to create the data directory, include the OS error. It will most likely be a permission denied error.

Lastly, a very minor log fixup in debug mode that was confusing to me when recording the hashes of new files in a downloaded tarball. Log them as being new instead of having a previous hash of "".

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -870,11 +870,11 @@ def test_suricata(suricata_path):
             test_command))
         env = {
             "SURICATA_PATH": suricata_path,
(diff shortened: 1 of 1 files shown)


victorjulien:
Tests OK here.

PR #64, opened 2018-11-22
Use setuptools to install requirements

shivan1b:
Current setup was using Python's distutils which does not allow setting
any required modules for the package. Change the flow of setup making it
use setuptools whenever possible to be able to install the requirements,
fallback to distutils otherwise. Warning about the option not recognized
by distutils has also been taken care of with this patch in case the
installation is done with distutils.
Now, suricata-update installs as a complete application taking care of
all the requirements.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2667

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,15 @@
 import os.path
 import subprocess
-from distutils.core import setup
 
 from suricata.update.version import version
+
+try:
+    from setuptools import setup
+    USE_SETUPTOOLS = True
+except ImportError:
+    from distutils.core import setup
+    USE_SETUPTOOLS = False
+
 
 def write_git_revision():
     if not os.path.exists(".git"):
@@ -17,13 +24,13 @@ def write_git_revision():
 
 write_git_revision()
 
-setup(
-    name="suricata-update",
-    version=version,
-    description="Suricata Update Tool",
-    author="Jason Ish",
-    author_email="ish@unx.ca",
-    packages=[
+args = {
+    "name": "suricata-update",
+    "version": version,
+    "description": "Suricata Update Tool",
+    "author": "Jason Ish",
+    "author_email": "ish@unx.ca",
+    "packages": [
         "suricata",
         "suricata.update",
         "suricata.update.commands",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -1,9 +1,16 @@
...
 from suricata.update.version import version
 
+try:
+    from setuptools import setup
+    USE_SETUPTOOLS = True
+except ImportError:
+    from distutils.core import setup
+    USE_SETUPTOOLS = False
+
+
```
jasonish: minor: extra line space.. Or is this style, 2 spaces after import block?

setup.py
```diff
@@ -32,12 +39,17 @@ def write_git_revision():
...
     ],
-    scripts = [
+    "scripts": [
         "bin/suricata-update",
     ],
-)
+}
+
+if USE_SETUPTOOLS:
+    args["install_requires"] = ["pyyaml", ]
```
jasonish: minor: extra space at end, or is this proper style again.

jasonish:
So while this works, and is exactly what I had in mind, I've found an interesting side affect.

If I have setuptools available, and py-yaml is not installed, it will go ahead and install it for me.  I'm not sure if we want this, especially when its wrapped by Suricata's make install.

Thought: Is there a way to NOT import setuptools, but instead detect if its running under pip, and then add install_requires if running under pip?

setup.py
shivan1b: Yes. We'll get E302 if this is not done. http://pycodestyle.pycqa.org/en/latest/intro.html

setup.py
shivan1b: Yes. E231 on the same doc.

shivan1b:
So, I found this one way of finding out if its running under pip or not:

```
if any("pip" in arg for arg in distutils.sys.argv):
     print("********************** Using pip ********************")
```

Does this look better? If yes, I could send a working PR with this right away.

jasonish:
> Does this look better? If yes, I could send a working PR with this right away.

Yup. This looks like it could work. So we'd always import distutils, and never setuptools, and then add the pyyaml dependency only if running under pip?  I think this could work well.

shivan1b:
Created #65

PR #65, opened 2018-11-23
Install requirements if using pip

shivan1b:
Current setup was using distutils which does not allow to define the
requirements of a package. Check if the installation of
`suricata-update` is being done with `pip` and if it is, install the
requirements while installing the package.
This way distutils will not throw a warning of the `install_requires`
option being unrecognized, however, it would still not install the
requirements.
Now, with the installation of `suricata-update` package, all the
requirements are installed as well if it is installed with `pip`.

This is a rework of #64 

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2667

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,10 @@
 import os.path
 import subprocess
+import distutils
 from distutils.core import setup
 
 from suricata.update.version import version
+
 
 def write_git_revision():
     if not os.path.exists(".git"):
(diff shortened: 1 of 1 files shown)


jasonish:
Merged with #66.  Thanks!

PR #67, opened 2018-11-28
Fix missing null check

shivan1b:
Coverity discovered a missing null check for the variable `val`. Fix
 that by adding the appropriate checks for it.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2676


jasonish:
Rebased and merged. Thanks!

PR #68, opened 2018-12-05
do not fail on repeated enabling/disabling of sources

satta:
At the moment, trying to enable a source that is already enabled will result in suricata-update failing with a return code of 1 and the error message:

```
2018-12-04 17:15:14,285 - <ERROR> - The source oisf/trafficid is already enabled.
```

This is problematic, for example, when using Ansible to set up and maintain an environment where configuration of sources is necessary. This PR introduces code that making this a warning only and not returning a nonzero return code.

  contribution agreement at

See ticket [2728](https://redmine.openinfosecfoundation.org/issues/2728).

Describe changes:
- Adapt messages for repeated calls of `suricata-update **enable-source`/`suricata-update disable-source` make them a warning, adapt wording.
- Ensure non-error return code when a source is already enabled/disabled.

### DIFF
--- a/suricata/update/commands/disablesource.py
+++ b/suricata/update/commands/disablesource.py
@@ -34,7 +34,7 @@ def disable_source():
     if not os.path.exists(filename):
         logger.debug("Filename %s does not exist.", filename)
         logger.warning("Source %s is not enabled.", name)
-        return 1
+        return 0
     logger.debug("Renaming %s to %s.disabled.", filename, filename)
     os.rename(filename, "%s.disabled" % (filename))
     logger.info("Source %s has been disabled", name)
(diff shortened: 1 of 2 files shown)


jasonish:
I agree that this make sense, that it should be an idempotent action and not fail out.  However, should it still do this when we enable source with a different parameter?  For example:
```
suricata-update enable-source et/pro secret-code=xxx
```
then I do
```
suricata-update enable-source et/pro secret-code=yyy
```
My feeling is that the best thing to do here would be to update the parameter with a log message, or error out requiring in this case the source be remove first.

satta:
> My feeling is that the best thing to do here would be to update the parameter with a log message, or error out requiring in this case the source be remove first.

Right. I would suggest doing the former, as I would consider that to feel most natural to me -- and would also allow the user to easily change the `secret-code` parameter (if there was a typo etc.) without having to disable it first.
If you agree I can update my PR to reflect that.

jasonish:
> If you agree I can update my PR to reflect that.

Agreed. I think this makes sense.

PR #69, opened 2018-12-06
Add the facility to mention headers while adding source

shivan1b:
Add header option to add source command

Facilitates addition of a header per source while adding the source.
Usage:
```
$ ./bin/suricata-update add-source --header "Auth: Basic"
```
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2577

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -351,6 +351,9 @@ class Fetch:
             "%s-%s" % (url_hash, self.url_basename(url)))
 
     def fetch(self, url):
+        net_arg = url
+        if isinstance(url, tuple):
+            url = url[0]
         tmp_filename = self.get_tmp_filename(url)
         if not config.args().force and os.path.exists(tmp_filename):
             if not config.args().now and \
@@ -368,7 +371,7 @@ class Fetch:
         try:
             tmp_fileobj = tempfile.NamedTemporaryFile()
             suricata.update.net.get(
-                url,
+                net_arg,
                 tmp_fileobj,
                 progress_hook=self.progress_hook)
             shutil.copyfile(tmp_fileobj.name, tmp_filename)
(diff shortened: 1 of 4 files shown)


suricata/update/net.py
```diff
@@ -111,25 +122,37 @@ def get(url, fileobj, progress_hook=None):
...
-    info = remote.info()
+    if isinstance(url, tuple):
+        header = url[1].split(":") if url[1] is not None else None
+        if header and is_header_clean(header=header):
+            logger.debug("Setting HTTP header %s to %s", header[0], header[1])
+            opener.addheaders = [(header[0], header[1]), ]
+        elif header:
+            logger.error("Header not set as it does not meet the criteria")
+        url = url[0]
+
```
jasonish: When using --header "X-Key: XXX", I noticed that the server actually receives 2 spaces between the `:` and the `XXX`. header[1] probably needs to be stripped here.

suricata/update/main.py
```diff
@@ -351,6 +351,9 @@ def get_tmp_filename(self, url):
             "%s-%s" % (url_hash, self.url_basename(url)))
 
     def fetch(self, url):
+        net_arg = url
+        if isinstance(url, tuple):
+            url = url[0]
```
jasonish: I'm not sure if I like the idea of a tuple here. It works, and perhaps we can leave it for now.  But what if we were to allow for http basic authentication as well, and we had to pass in a username and password?  Would you do you this differently?

I'm pretty sure http basic auth will come as a feature request sooner or later, so I think its a good example for future growth in the arguments passed here.

jasonish: Yes, that would be the case.

PR #70, opened 2018-12-13
version: hide git revision if unknown - v1

jasonish:
This happens when suricata-update is installed bundled with Suricata.  This is more a relic from the earlier days of Suricata-Update when more users were likely to install it directly from git anyways.

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -67,7 +67,7 @@ from suricata.update.version import version
 try:
     from suricata.update.revision import revision
 except:
-    revision = "unknown"
+    revision = None
 
 # Initialize logging, use colour if on a tty.
 if len(logging.root.handlers) == 0 and os.isatty(sys.stderr.fileno()):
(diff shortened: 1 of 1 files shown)


suricata/update/main.py
```diff
@@ -1052,7 +1052,11 @@ def _main():
     global_args, rem = global_parser.parse_known_args()
 
     if global_args.version:
-        print("suricata-update version %s (rev: %s)" % (version, revision))
+        if revision != None:
+            revision_string = " (rev: %s)" % (revision)
+        else:
+            revision_string = ""
```
shivan1b: Minor: We could do these four lines in a one liner:

`revision_string = " (rev: %s)" % (revision) if revision else ""`

suricata/update/main.py
```diff
@@ -1052,7 +1052,11 @@ def _main():
...
+        print("suricata-update version %s%s" % (version, revision_string))
```
shivan1b: We should also probably start using `string.format` as `%` is the older way of string formatting. See: https://docs.python.org/3.7/library/string.html#format-examples

Works since 2.7 (backported).

So we should instead have:
`print("suricata-update version {}{}".format(version, revision_string))`

jasonish:
Thanks. Update here: https://github.com/OISF/suricata-update/pull/74

PR #72, opened 2018-12-13
Add the facility to mention headers while adding source

shivan1b:
Add header option to add source command

Facilitates addition of a header per source while adding the source.
Usage:

$ ./bin/suricata-update add-source --header "Auth: Basic"
Link to redmine ticket: https://redmine.openinfosecfoundation.org/issues/2577

This is a rework of #71 
Change: Updated Error message

### DIFF
--- a/suricata/update/commands/addsource.py
+++ b/suricata/update/commands/addsource.py
@@ -37,7 +37,6 @@ def register(parser):
 
 def add_source():
     args = config.args()
-    header = None
 
     if args.name:
         name = args.name
@@ -59,8 +58,7 @@ def add_source():
             if url:
                 break
 
-    if args.header:
-        header = args.header
+    header = args.header if args.header else None
 
     source_config = sources.SourceConfiguration(name, header=header, url=url)
     sources.save_source_config(source_config)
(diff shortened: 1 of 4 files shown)


suricata/update/commands/addsource.py
```diff
@@ -32,6 +32,7 @@ def register(parser):
     parser.add_argument("name", metavar="<name>", nargs="?",
                         help="Name of source")
     parser.add_argument("url", metavar="<url>", nargs="?", help="Source URL")
+    parser.add_argument("--header", metavar="<header>", help="HTTP Header")
```
jasonish: After using this command, I think this would be better named `--http-header` with a help string like "Additional HTTP header to add to requests"

suricata/update/net.py
```diff
@@ -111,25 +122,38 @@ def get(url, fileobj, progress_hook=None):
...
         opener.addheaders = [(header, value) for header,
                              value in opener.addheaders if header.lower() != "user-agent"]
-    remote = opener.open(url)
-    info = remote.info()
+    if isinstance(url, tuple):
+        header = url[1].split(":") if url[1] is not None else None
+        if header and is_header_clean(header=header):
+            name, val = header[0].strip(), header[1].strip()
+            logger.debug("Setting HTTP header %s to %s", name, val)
+            opener.addheaders = [(name, val), ]
```
jasonish: Just noticed this after some more testing - this appears to erase the user header set above.  Probably need to construct the list and then set it once before making the request.

shivan1b:
No problem. Addressing these issues in a new PR.

PR #83, opened 2018-12-21
Make header an optional argument

shivan1b:
header was being used as a required argument in SourceConfiguration
class. Since SourceConfiguration class has been used multiple times
across commands, not all the commands require header argument and thus
it breaks. Fix it by making header an optional argument.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2753

### DIFF
--- a/suricata/update/sources.py
+++ b/suricata/update/sources.py
@@ -77,7 +77,7 @@ def save_source_config(source_config):
 
 class SourceConfiguration:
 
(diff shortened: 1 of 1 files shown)


jasonish:
Thanks! Merged into master-1.0.x and released in 1.0.3, and merge into master.

PR #85, opened 2018-12-22
do not fail on repeated enabling/disabling of sources -- v2

satta:
Previous PR: https://github.com/OISF/suricata-update/pull/68

Makes sure to update parameters when re-enabling a source, printing a log message for each changed parameter.

  contribution agreement at
- [X] I have updated the user guide (in doc/) to reflect the
  changes made 

See ticket [2728](https://redmine.openinfosecfoundation.org/issues/2728).

Describe changes:
- Do not simply return when source is already enabled or re-enabled, make sure to always write a new source config. Initialize with existing configuration if so.
- Print `info` level log message when updating a value in the existing configuration.
- Mention the update behavior in the documentation for `enable-source`.

### DIFF
--- a/suricata/update/commands/enablesource.py
+++ b/suricata/update/commands/enablesource.py
@@ -40,20 +40,21 @@ def register(parser):
 
 def enable_source():
     name = config.args().name
+    updating = False
 
     # Check if source is already enabled.
     enabled_source_filename = sources.get_enabled_source_filename(name)
     if os.path.exists(enabled_source_filename):
-        logger.warning("The source %s is already enabled, keeping it enabled.", name)
-        return 0
+        logger.warning("The source %s is already enabled.", name)
+        updating = True
 
     # First check if this source was previous disabled and then just
     # re-enable it.
     disabled_source_filename = sources.get_disabled_source_filename(name)
     if os.path.exists(disabled_source_filename):
-        logger.info("Re-enabling previous disabled source for %s.", name)
+        logger.info("Re-enabling previously disabled source for %s.", name)
         os.rename(disabled_source_filename, enabled_source_filename)
-        return 0
+        updating = True
 
     if not os.path.exists(sources.get_index_filename()):
         logger.warning("Source index does not exist, will use bundled one.")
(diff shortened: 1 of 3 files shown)


suricata/update/commands/enablesource.py
```diff
@@ -71,25 +72,35 @@ def enable_source():
...
+                if param in opts:
+                    params[param] = opts[param]
+                else:
+                    prompt = source["parameters"][param]["prompt"]
+                    while True:
+                        r = input("%s (%s): " % (prompt, param))
+                        r = r.strip()
+                        if r:
+                            break
+                    params[param] = r.strip()
```
shivan1b: Since not all of the code in the above added block is newly added or modified, we could probably do it in a cleaner way following _Flat is better than nested._ from the [Zen of Python](https://www.python.org/dev/peps/pep-0020). ;)

So, something like, leave the code that is already there and do

```
+    if updating:
+        source = yaml.safe_load(open(sources.get_enabled_source_filename(name), "rb"))
+    else:
+        source = source_index.get_sources()[name]
+
+    if "params" in source:
+        params = source["params"]
+        for old_param in source["params"]:
+            if old_param in opts and source["params"][old_param] != opts[old_param]:
+                logger.info("Updating source parameter '%s': '%s' -> '%s'." % (
+                    old_param, source["params"][old_param], opts[old_param]))
+                params[old_param] = opts[old_param]
```

Will it make sense or mess up with your logic?

shivan1b:
Hey @satta !
Thanks for your first contribution to the project. :)

suricata/update/commands/enablesource.py
```diff
@@ -40,20 +40,21 @@ def register(parser):
 
 def enable_source():
     name = config.args().name
+    updating = False
```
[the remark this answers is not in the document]
satta: Sure.

suricata/update/commands/enablesource.py
satta: I think this will work. I have also tested the changed code a bit in various situations. I'll address this and send a new PR. Thanks for taking a look!

PR #87, opened 2019-01-10
Rules log file/v3

shivan1b:
Add a `--report <filename>` option so that the end user is able to log a
report about the rules enabled, disabled, modified and dropped.

Usage
```
└─ $ ▶ ./bin/suricata-update --report /tmp/rules.log
```

Sample Report

```
Generated on Wednesday, 09 Jan 2019, 23:30:43

Added Rules
===========
[1:2008402] ET MALWARE Realtimegaming.com Online Casino Spyware Gaming Checkin
(emerging-malware.rules)
(emerging-malware.rules)
(emerging-malware.rules)
(emerging-malware.rules)
```

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2256

NOTE: Kindly ignore the formatting here, please check on command line.

Changes since v2:
- Added filename for rules
- Changed the format of report
- Used rule.brief for clearer logging

This is a rework of #86

PR #88, opened 2019-01-10
Rules log file/v4

shivan1b:
Add a `--report <filename>` option so that the end user is able to log a
report about the rules enabled, disabled, modified and dropped.

Usage
```
└─ $ ▶ ./bin/suricata-update --report /tmp/rules.log
```

Sample Report

```
Generated on Wednesday, 09 Jan 2019, 23:30:43

Added Rules
===========
 # [1:2002101] ET GAMES Battle.net Starcraft login
(emerging-games.rules)
 # [1:2002102] ET GAMES Battle.net Brood War login
(emerging-games.rules)
 # [1:2002103] ET GAMES Battle.net Diablo login
(emerging-games.rules)
 # [1:2002104] ET GAMES Battle.net Diablo 2 login
(emerging-games.rules)
```

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2256

Changes since v3:
- Fix the commit message

This is a rework of #87

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -32,6 +32,8 @@ import shutil
 import glob
 import io
 import tempfile
+import json
+from datetime import datetime as dt
 
 try:
     # Python 3.
@@ -68,7 +70,6 @@ try:
     from suricata.update.revision import revision
 except:
     revision = None
-
 # Initialize logging, use colour if on a tty.
 if len(logging.root.handlers) == 0 and os.isatty(sys.stderr.fileno()):
     logger = logging.getLogger()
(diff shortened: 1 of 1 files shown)


jasonish:
1) Can we get this on one line:
```
# [1:2002101] ET GAMES Battle.net Starcraft login
(emerging-games.rules)
```
I wonder if it would be worth dropping the '.rules' to shorten it up a bit. Probably doesn't matter.

2) I think we should have a summary as the first section:
```
Summary
========
Added: 12
Removed: 13
Modified: 99
```
this at least gives us something with all 0's when nothing has changed which is better than an empty report.

This is also where it might be nice to add extra information like what Suricata-Update itself did:
```
Rules enabled for flowbit dependencies: 188
Rules disabled by disable.conf: 12
Rules enabled by enable.conf: 13
Rules converted to drop: 9
Rules modified by modify.conf: 0
```
This should be a separate commit after the first piece, as the first piece is already a functioning unit of code. Just needs a minor fixup from (1).

shivan1b:
Please have a look at #91

PR #90, opened 2019-01-14
Avoid failure in case of obsoletion of attribute

shivan1b:
`platform` is currently (Py3.7) shipped with `linux_distribution`
attribute but it has been Deprecated since Py3.5. In future when it
becomes obsolete, suricata-update will break while creating User-Agent
header. In order to avoid this, check for the attribute before adding it
to User-Agent.

This renders redmine issue #2645 invalid.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2645

### DIFF
--- a/suricata/update/net.py
+++ b/suricata/update/net.py
@@ -67,7 +67,7 @@ def build_user_agent():
     params.append("CPU: %s" % (platform.machine()))
     params.append("Python: %s" % (platform.python_version()))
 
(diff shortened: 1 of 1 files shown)


jasonish:
Merged.  Thanks Shivani!

PR #94, opened 2019-01-22
Add timeout of 30 sec for all GET requests

shivan1b:
suricata-update has been reported to get hung in case the download fails
for a particular source. Add a timeout parameter to urllib to avoid that
and continue further processing after a timeout of 30 seconds.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2703


shivan1b:
Merged into master.

PR #95, opened 2019-02-01
Issue 2791 - optimize memory usage - v1

jasonish:
Link
to
https://redmine.openinfosecfoundation.org/issues/2791

These PR stops recording all the rule options into a list.  That is left over from idstools that have some library support for modifying rules that is not required by Suricata  updating. In my tests, removing this improved memory usage by 10-20%.


jasonish:
See https://github.com/OISF/suricata-update/pull/96

PR #96, opened 2019-02-11
Issue 2791 - optimize memory usage - v2

jasonish:
Redmine issue:
https://redmine.openinfosecfoundation.org/issues/2791

Previous PR:
- https://github.com/OISF/suricata-update/pull/95

Changes from last PR:
- When building the mini-report of rule updates, only track IDs when we can to limit the amount of memory used.
- Set some collections to None to free memory before starting up Suricata in test mode.

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -566,75 +566,76 @@ def load_dist_rules(files):
                 logger.error("Failed to open %s: %s" % (path, err))
                 sys.exit(1)
 
-def build_report(prev_rulemap, rulemap):
-    """Build a report of changes between 2 rulemaps.
-
-    Returns a dict with the following keys that each contain a list of
-    rules.
-    - added
-    - removed
-    - modified
-    """
-    report = {
-        "added": [],
-        "removed": [],
-        "modified": []
-    }
-
-    for key in rulemap:
-        rule = rulemap[key]
-        if not rule.id in prev_rulemap:
-            report["added"].append(rule)
-        elif rule.format() != prev_rulemap[rule.id].format():
-            report["modified"].append(rule)
-    for key in prev_rulemap:
-        rule = prev_rulemap[key]
-        if not rule.id in rulemap:
-            report["removed"].append(rule)
-
-    return report
-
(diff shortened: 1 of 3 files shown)


victorjulien:
With py2 this works for me, but when forcing py3, I get:
<pre>
11/2/2019 -- 17:56:56 - <Info> -- Loading local file /home/victor/dev/suricata-trafficid/rules/traffic-id.rules
Traceback (most recent call last):
  File "/home/victor/dev/suricata-update/bin/suricata-update", line 33, in <module>
    sys.exit(main.main())
  File "/home/victor/dev/suricata-update/suricata/update/main.py", line 1457, in main
    sys.exit(_main())
  File "/home/victor/dev/suricata-update/suricata/update/main.py", line 1329, in _main
    rulemap = build_rule_map(rules)
  File "/home/victor/dev/suricata-update/suricata/update/main.py", line 694, in build_rule_map
    if rule["rev"] > rulemap[rule.id]["rev"]:
TypeError: '>' not supported between instances of 'NoneType' and 'NoneType'
</pre>

Cmdline: "# su suricata -c "python3 /home/victor/dev/suricata-update/bin/suricata-update"

jasonish:
Interesting. Py2 allows > between 2 none types and Py3 doesn't.  I'll update suricata-update to design a default rev of 0 if not provided by the rule.

jasonish:
Python 3 issue fixed in new PR: https://github.com/OISF/suricata-update/pull/97

victorjulien:
Any idea why travis didn't catch this?

jasonish:
The tests don't download the trafficid tests, and all other rulesets include a rev in every rule.  While I fixed it here, I should also check why trafficid doesn't have a rev on some rules.

jasonish:
I should also note that the tests run on Travis are crafted to not hit the network. We could change that.

victorjulien:
Could we bundle some test rule files?

jasonish:
We do, but as mentioned, they all have the rev.  So we'll do a new test that checks this.  Will, the test for now can just include to test that rules without a rev get assigned the default.

PR #98, opened 2019-02-22
Fix null pointer dereference issue detected by coverity

shivan1b:
Coverity scan of the current code raised the following issue:
```
>>>     CID 327298:  Null pointer dereferences  (FORWARD_NULL)
>>>     Accessing a property of null-like value "val".
271                 if val.find("noalert") > -1:
```

Fix this by adding a null check on val.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2834


shivan1b:
Merged into master.

PR #103, opened 2019-03-12

vrindaaa:

- [Done ] I have read the contributing guide lines at
- [ Done] I have signed the Open Information Security Foundation
  contribution agreement at
- [Not Applicable ] I have updated the user guide (in doc/userguide/) to reflect the
  changes made (if applicable)

Describe changes:
- This issue can be fixed by adding a check to make sure before converting them to int that self.gid and self.sid are not NoneType.

### DIFF
--- a/suricata/update/rule.py
+++ b/suricata/update/rule.py
@@ -118,7 +118,13 @@ class Rule(dict):
         :returns: A tuple (gid, sid) representing the ID of the rule
         :rtype: A tuple of 2 ints
         """
-        return (int(self.gid), int(self.sid))
+        self_gid=self.gid
+        self_sid=self.sid
+        if self.gid is not None:
+            self_gid=int(self_gid)
+        if self.sid is not None:
+            self_sid=int(self_sid)
+        return (self_gid, self_sid)
 
     @property
     def idstr(self):
--- a/suricata/update/rule.py
+++ b/suricata/update/rule.py
@@ -118,13 +118,7 @@ class Rule(dict):
         :returns: A tuple (gid, sid) representing the ID of the rule
         :rtype: A tuple of 2 ints
         """
-        self_gid=self.gid
-        self_sid=self.sid
-        if self.gid is not None:
-            self_gid=int(self_gid)
-        if self.sid is not None:
-            self_sid=int(self_sid)
-        return (self_gid, self_sid)
+        return (int(self.gid), int(self.sid))
 
     @property
     def idstr(self):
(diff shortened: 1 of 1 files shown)


suricata/update/rule.py
```diff
@@ -118,7 +118,14 @@ def id(self):
...
         """
-        return (int(self.gid), int(self.sid))
+        
+        self_gid=self.gid
+        self_sid=self.sid
+        if self.gid is not None:
+            self_gid=int(self_gid)
+        if self.sid is not None:
+            self_sid=int(self_sid)
+        return (self_gid, self_sid)
```
shivan1b: This is incorrect. `sid` and `gid` fields are pretty improtant for any rule and are never set to `NoneType` after parsing. You should instead handle the exception cleanly so that in case of failure there is no ugly traceback.

shivan1b:
@vrindaaa thanks for your contribution. Please address the requested changes. Apart from that,
- Clean your commit history, your first commit was not needed at all
- Do not prefix `Subject:` to the commit heading
- Make sure your commit messages are descriptive
- In cases like this issue, it would be good to include test results in commit message showing what it looked like before your changes and after them
- Wrap your commit messages to 80 char length

vrindaaa:
@shivan1b thank you for your feedback! I have tried to work on the things you said in the latest commit.
Please look into the same and tell me if any further changes need to be made.
I am new to open source so I really appreciate all of your help!

shivan1b:
@vrindaaa I'm afraid its still not correct. BTW, following the contribution guidelines (the link shared above) you should create a new PR for every new change. Could you please read the contribution guidelines here and on Outreachy project page for suricata-update once again?
Apart from this, the old unnecessary commit is still there. :(

suricata/update/rule.py
```diff
@@ -154,6 +154,10 @@ def find_opt_end(options):
 
     while True:
         i = options[offset:].find(";")
+        if(options[offset:i+1].find("classtype")!=-1):
+            j=options[offset+2:].find(" ")
+            if(j!=-1):
+                i=j
```
shivan1b: When I gave an example about the missing semicolon after `classtype` var, I did not mean that it will be the only case. A user can miss a semicolon anywhere in the rule.
To be clear, we need to catch the Exception that is caused in such a case and exit with a clean error instead of an ugly traceback.
Is this helpful?

vrindaaa:
So if I want to remove all my previous commits and make a new commit for this issue (my new changes are completely unrelated to the previous ones, and I will undo those), should I make another PR or can I just delete my previous commits and make the new commit here?

shivan1b:
You should alwayd for even a smallest change create a new PR and close the old one.

PR #107, opened 2019-03-13
Optimization : Catch control-c (KeyboardInterrupt) and exit cleanly. #4

vrindaaa:
Optimization #2878
Catch control-c (KeyboardInterrupt) and exit cleanly.

Can be done by importing python library signal, and defining a function
signal_handler which will print a message if keyboard interrupt is
detected and exit the program.

Checklist:

contribution agreement at
- [ ]  I have updated the user guide (in doc/userguide/) to reflect the
changes made (if applicable)

Link to redmine :
https://redmine.openinfosecfoundation.org/issues/2878

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -32,6 +32,7 @@ import shutil
 import glob
 import io
 import tempfile
+import signal
 
 try:
     # Python 3.
@@ -1452,7 +1453,12 @@ def _main():
 
     return 0
 
+def signal_handler(signal, frame):
+    print('Program interrupted. Aborting...')
+    sys.exit(1)
+
 def main():
+    signal.signal(signal.SIGINT, signal_handler)
     try:
         sys.exit(_main())
     except exceptions.ApplicationError as err:


vrindaaa:
Thank you for all your help!

jasonish:
While this seems to work fine, I'm curious about the choice to use a signal handler instead of catching a KeyboardInterrupt?

sh0213:
Apparently I missed that as it wasn't marked on the ticket but I did send a PR using a `KeyboardInterrupt` funnily enough: https://github.com/OISF/suricata-update/pull/117

vrindaaa:
Initially I had made a pull request which was catching the KeyboardInterrupt, though that introduced an extra level of nesting. After that I was trying to find alternate ways to do the same and signal handler seemed to work fine. Is catching a KeyboardInterrupt a better option?

shivan1b:
@vrindaaa the right way to have done it with the `KeyboardInterrupt` is as done in #117 extra level of nesting was not required.
BTW, @jasonish I found this signal approach cleaner and more readable. This is also going to work on all the platforms and not affect the existing code. I had also read that catching KeyboardInterrupt does not take into account the signals by `kill` command for abrupt killing with SIGINT but I just tested and seems like it does work.
 `signal` package also does not require any external installations.
Do you have a different opinion?

jasonish:
My initial thought was that it was going to work, but SIGINT, Ctrl-C is supported. Just checked, the functioned registered for atexit fires as well.  I just thought that:
```
    try: 
        sys.exit(_main())
    except exceptions.ApplicationError as err: 
        logger.error(err)
    except KeyboardInterrupt:
        logger.info("Program interrupted. Aborting...")
    sys.exit(1)
```
might be more idiomatic. But its OK with me as-is.

shivan1b:
Merged into master. Thanks for your contribution. :)

PR #108, opened 2019-03-13
Fix TypeError : Failure with missing rule params #2

vrindaaa:

To correct this I added a try-except statement in rule.py which would
raise an error if sid or gid are of type None because of incorrect
parsing of a rule.

contribution agreement at
changes made (if applicable)

### DIFF
--- a/suricata/update/rule.py
+++ b/suricata/update/rule.py
@@ -118,7 +118,11 @@ class Rule(dict):
         :returns: A tuple (gid, sid) representing the ID of the rule
         :rtype: A tuple of 2 ints
         """
-        return (int(self.gid), int(self.sid))
+        try:
+            return (int(self.gid), int(self.sid))
+        except TypeError:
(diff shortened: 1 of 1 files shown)


suricata/update/rule.py
```diff
@@ -118,7 +118,11 @@ def id(self):
         :returns: A tuple (gid, sid) representing the ID of the rule
         :rtype: A tuple of 2 ints
         """
-        return (int(self.gid), int(self.sid))
+        try:
+            return (int(self.gid), int(self.sid))
+        except TypeError:
+            print("Error : Sid cannot be null")
```
shivan1b: It would probably make more sense to use logger to log this error. Also, since we do not know whether sid or gid failed, we should make the error message clearer.

shivan1b:
@vrindaaa apart from the changes requested in code, please make sure to use imperative mood in your sentences for entire commit message.

PR #110, opened 2019-03-13
Fix TypeError : Failure with missing rule params #3

vrindaaa:

To correct this I added a try-except statement in rule.py which would
raise an error if sid or gid are of type None because of incorrect
parsing of a rule.

contribution agreement at
- [ ]  I have updated the user guide (in doc/userguide/) to reflect the
changes made (if applicable)

### DIFF
--- a/suricata/update/rule.py
+++ b/suricata/update/rule.py
@@ -118,7 +118,11 @@ class Rule(dict):
         :returns: A tuple (gid, sid) representing the ID of the rule
         :rtype: A tuple of 2 ints
         """
-        return (int(self.gid), int(self.sid))
+        try:
+            return (int(self.gid), int(self.sid))
+        except TypeError:
+            logger.error("sid or gid can't be null, exiting.")
+            sys.exit(1)
 
     @property
     def idstr(self):


vrindaaa:
Heyy, @shivan1b can you please look this over once?

shivan1b:
@jasonish could you please have a look at this? Do you think this is OK?

suricata/update/rule.py
```diff
@@ -118,7 +118,11 @@ def id(self):
         :returns: A tuple (gid, sid) representing the ID of the rule
         :rtype: A tuple of 2 ints
         """
-        return (int(self.gid), int(self.sid))
+        try:
+            return (int(self.gid), int(self.sid))
+        except TypeError:
+            logger.error("sid or gid can't be null, exiting.")
+            sys.exit(1)
```
jasonish: I don't think we should exit here. The code in this file should be considered library code, and not cause the program to exit. I think the better fix is to return an error that is more descriptive of what is going on, like an exception with a message that rule has no sid, and/or perhaps an invalid sid (it can't be parsed as an integer).  Then the caller can decide what to do.

I'd rather see the program continue running and log this error rather than abort.

vrindaaa: Can you please explain what the function should exactly return, if not exit. The values it is returning are being used in the main.py file so if it doesn't return a tuple the main code would show error. One possible solution I could think of was to log the error and ask the user if they want to ignore this rule (not consider it as it does not have a valid sid) or input it's sid. So should I go with that?

jasonish: After thinking about this more, and seeing that it is used as a property, I think we need to set it up to *never* fail.  Or leave it as is, so it fails with a full proper traceback.  So I don't think this method should change at all.

Instead we should probably validate the rule during parsing, before we return this Rule class.

I might do something like create a "BadSidError" exception.  And near the end of rule.parse(), validate that the sid is a integer, and if not raise a BadSidError exception.  Generally the rule parsing code is going to more tolerant to this error than code that just wants to print the rule.id.

PR #113, opened 2019-03-15
Fix TypeError : Failure with missing rule params #4

vrindaaa:

contribution agreement at
changes made (if applicable)

### DIFF
--- a/suricata/update/rule.py
+++ b/suricata/update/rule.py
@@ -286,6 +286,9 @@ def parse(buf, group=None):
 
     return rule
 
+class BadSidError(Exception):
+    """Raises exception when sid is of type null"""
+
 def parse_fileobj(fileobj, group=None):
     """ Parse multiple rules from a file like object.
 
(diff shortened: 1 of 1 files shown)


vrindaaa:
@jasonish here, the check which has failed is the one where the rule didn't have a sid so I didn't append it to the final rules list. Should this problem be approached in a different way?

suricata/update/rule.py
```diff
@@ -310,7 +313,13 @@ def parse_fileobj(fileobj, group=None):
...
             rule = parse(buf, group)
             if rule:
-                rules.append(rule)
+                try:
+                    if rule["sid"]!=None:
+                        rules.append(rule)
+                    else:
+                        raise BadSidError("Sid cannot be of type null")
+                except BadSidError as err:
+                    logger.error("Failed to parse rule: %s: %s", buf.rstrip(), err)
```
shivan1b: Was this extra nested try..except block required? Could you please try doing it in a cleaner way?

suricata/update/rule.py
```diff
@@ -310,7 +313,13 @@ def parse_fileobj(fileobj, group=None):
         try:
             rule = parse(buf, group)
             if rule:
-                rules.append(rule)
+                try:
+                    if rule["sid"]!=None:
```
shivan1b: Please follow PEP8 standards.

shivan1b: Is `!= None` check required?

shivan1b:
@vrindaaa seems like this would require some fixes to the tests as well. As you can see some tests are failing which do not have `sid` in them. Could you please borrow the rules from other tests that have rules that include `sid` and use them in these tests?
Thanks.

PR #118, opened 2019-03-20
Fix TypeError : Failure with missing rule params #5

vrindaaa:

contribution agreement at
changes made (if applicable)

### DIFF
--- a/suricata/update/rule.py
+++ b/suricata/update/rule.py
@@ -286,6 +286,9 @@ def parse(buf, group=None):
 
     return rule
 
+class BadSidError(Exception):
+    """Raises exception when sid is of type null"""
+
 def parse_fileobj(fileobj, group=None):
     """ Parse multiple rules from a file like object.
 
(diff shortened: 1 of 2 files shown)


suricata/update/rule.py
```diff
@@ -310,7 +313,12 @@ def parse_fileobj(fileobj, group=None):
         try:
             rule = parse(buf, group)
             if rule:
-                rules.append(rule)
+                try:
+                    if rule["sid"]==None:
```
shivan1b: PEP8 standards still not followed. Spaces around operators. Again, was this `==None` necessary?

suricata/update/rule.py
```diff
@@ -310,7 +313,12 @@ def parse_fileobj(fileobj, group=None):
...
+                        raise BadSidError("Sid cannot be of type null")
+                    rules.append(rule)
+                except BadSidError as err:
```
shivan1b: This try except block is redundant.

shivan1b:
@vrindaaa based on the editor you're using, you should probably get a plugin for PEP8. Otherwise, have a look at this module here: https://pypi.org/project/autopep8/

PR #124, opened 2019-03-23
Bring parsers out of the main function

vidushi4:
In an attempt to make the main function code cleaner and compact, separate the parsers out by defining a function parse_args(). Since the args variable is a global variable, the simplest approach is to define a new function and make a function call to it from main.

Closes Redmine ticket, Optimization #2874

-- thank you.

  contribution agreement at
  changes made (if applicable)

Link
to
[https://redmine.openinfosecfoundation.org/issues/2874](url)
Describe changes:
Define a new function parse_args() to bring the parsers out of main function and make a call from main function.

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -1017,8 +1017,8 @@ def check_output_directory(output_dir):
             raise exceptions.ApplicationError(
                 "Failed to create directory %s: %s" % (
                     output_dir, err))
-
-def _main():
+           
+def parse_args():
     global args
 
     default_update_yaml = config.DEFAULT_UPDATE_YAML_PATH
@@ -1179,7 +1179,10 @@ def _main():
             setattr(args, arg, getattr(global_args, arg))
         elif hasattr(args, arg) and getattr(args, arg) is None:
             setattr(args, arg, getattr(global_args, arg))
-
+    
+def _main():
+    parse_args()
+    
     # Go verbose or quiet sooner than later.
     if args.verbose:
         logger.setLevel(logging.DEBUG)


suricata/update/main.py
```diff
@@ -1179,7 +1179,10 @@ def _main():
             setattr(args, arg, getattr(global_args, arg))
         elif hasattr(args, arg) and getattr(args, arg) is None:
             setattr(args, arg, getattr(global_args, arg))
-
+    
+def _main():
+    parse_args()
+
```
shivan1b: There is a lot of extra whitespace in this. Please configure your git to show that before committing. Can we also move parsers to a parsers.py in the same directory?
@jasonish  will it be OK to make a parsers.py and move parser code there?

jasonish: Which parsers? The command line parsers?

vidushi4: Yes, the command line parsers are to be separated out, right @shivan1b ? 
Should I go ahead and try moving parsers to a new file parsers.py?

shivan1b: @jasonish yes. Since we add a lot of subparsers and options, it makes the main function too big, so bringing all of it out of the main. Do you think making a parsers.py and adding them there would be a good idea? We could separate the parsing logic and the main logic with this.

@vidushi4 please wait for Jason's opinions about the same.

jasonish: > @jasonish yes. Since we add a lot of subparsers and options, it makes the main function too big, so bringing all of it out of the main. Do you think making a parsers.py and adding them there would be a good idea? We could separate the parsing logic and the main logic with this.

Its trick.  The main only contains 2 parser, and only ever should I think.  Thats the global options and the ones for the default command.  Perhaps just moving them to a method.

Every other subcommand sets up its own parser, so those should probably stay with the command that they are related to, so I wouldn't want to see those broken out either.

So refactor into one of more functions for sure.  Not to keen on it being in another file, IMO.

shivan1b: Ah right. Totally overlooked the `global` options. Thanks!
@vidushi4 please create a new PR with the changes as requested by Jason in this comment above.

vidushi4: Okay, so I should make two functions for the two parsers and let the parsers of subcommands be in main with the command they are related to, right?

shivan1b: @vidushi4 create a `parsers.py` and move all the parser code that is not affected by the global variables there. Create one separate function within `main.py` to deal with the functionality that involves using the global variables.
This way we make `_main()` as well as `main.py` cleaner without breaking anything.
Does this make sense to you?

vidushi4: Thanks for the clarification! I'll create a new PR soon.

vidushi4: There are two parsers, `global_parser` is the parent parser for the subparsers. Thus, rest of the parser code is affected by `global_parser`. I should ideally move both the parsers, global (parent parser) and its subparsers, to `parsers.py` to not let it be broken. 
I'm sort of confused and still not quite clear with the approach. @shivan1b could you please help me out here?
Also, Jason had said "So refactor into one of more functions for sure. Not too keen on it being in another file". Are we still going with making another file?

shivan1b: What you say sounds about right. Have you tried doing that and testing the code?
Re Jason's take on another file: Yes. I explained to him what I am looking forward to. He's fine with the new file as long as things don't break.

vidushi4: Yes, I moved both global and subparsers in a new file `parsers.py` and I've tested locally, it seems to work fine. Should I send a PR of the same?

vidushi4: Hi @shivan1b , should I close this PR and open a fresh PR with a new file `parsers.py` containing both parent and subparsers which I've tested locally?

shivan1b: Yes, please.

vidushi4:
Replaced by #143

PR #127, opened 2019-03-26

vrindaaa:

contribution agreement at
changes made (if applicable)


shivan1b:
@vrindaaa your commit message could be imperative. Like the heading of your PR.

tests/test_rule.py
```diff
@@ -81,8 +81,8 @@ def test_parse_fileobj(self):
         self.assertEqual(2, len(rules))
 
     def test_parse_file(self):
```
shivan1b: We should also have a test for the case you just covered. Could you try and add that?

vrindaaa: How can I run these tests on my laptop?

PR #129, opened 2019-03-26

vrindaaa:

contribution agreement at
changes made (if applicable)

### DIFF
--- a/suricata/update/rule.py
+++ b/suricata/update/rule.py
@@ -286,6 +286,9 @@ def parse(buf, group=None):
 
     return rule
 
+class BadSidError(Exception):
+    """Raises exception when sid is of type null"""
+
 def parse_fileobj(fileobj, group=None):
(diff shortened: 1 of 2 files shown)


vrindaaa:
@shivan1b could you explain to me why these checks would fail?

shivan1b:
@vrindaaa seems like plain incorrect to me. Could you please tell why do you think it should not have failed?

vrindaaa:
@shivan1b I figured out my mistake thanks. Do you want me to create a seperate test to check it or make changes in an existing test? Also could you help me with running these tests on my laptop?

shivan1b:
I think change in the existing test is fine since the tests are on per function basis.

On your laptop, install `tox` and from the root of suricata-update, just run `tox`. It will run all the tests on all the Python versions mentioned in the configuration file and available in your system.

PR #130, opened 2019-03-26

vrindaaa:

contribution agreement at
changes made (if applicable)

### DIFF
--- a/tests/test_rule.py
+++ b/tests/test_rule.py
@@ -71,8 +71,8 @@ class RuleTestCase(unittest.TestCase):
         self.assertEqual(str(rule), """alert tcp $HOME_NET any -> $EXTERNAL_NET any (msg:"some message";)""")
 
     def test_parse_fileobj(self):
-        rule_buf = ("""# alert tcp $HOME_NET any -> $EXTERNAL_NET any """
-                    """(msg:"some message";)""")
+        rule_buf = ("""alert ( msg:"DECODE_NOT_IPV4_DGRAM"; sid:1; gid:116; rev:1; metadata:rule-type decode; classtype:protocol-command-decode;) """
+                    """alert ( msg:"DECODE_NOT_IPV4_DGRAM"; sid:1; gid:116; rev:1; metadata:rule-type decode; classtype:protocol-command-decode;)""")
         fileobj = io.StringIO()
         for i in range(2):
             fileobj.write(u"%s\n" % rule_buf)
@@ -81,14 +81,24 @@ class RuleTestCase(unittest.TestCase):
         self.assertEqual(2, len(rules))
 
     def test_parse_file(self):
-        rule_buf = ("""# alert tcp $HOME_NET any -> $EXTERNAL_NET any """
-                    """(msg:"some message";)""")
+        rule_buf = ("""alert ( msg:"DECODE_NOT_IPV4_DGRAM"; sid:1; gid:116; rev:1; metadata:rule-type decode; classtype:protocol-command-decode;) """
+                    """alert ( msg:"DECODE_NOT_IPV4_DGRAM"; sid:1; gid:116; rev:1; metadata:rule-type decode; classtype:protocol-command-decode;)""")
         tmp = tempfile.NamedTemporaryFile()
         for i in range(2):
             tmp.write(("%s\n" % rule_buf).encode())
         tmp.flush()
         rules = suricata.update.rule.parse_file(tmp.name)
         self.assertEqual(2, len(rules))
+
+    def test_parse_rule_with_sid_error(self):
+        rule_buf = ("""# alert tcp $HOME_NET any -> $EXTERNAL_NET any (msg:"some message";)"""
+                    """alert ( msg:"DECODE_NOT_IPV4_DGRAM" sid:1; gid:116; rev:1; metadata:rule-type decode; classtype:protocol-command-decode;)""")
+        fileobj = io.StringIO()
+        for i in range(2):
+            fileobj.write(u"%s\n" % rule_buf)
+        fileobj.seek(0)
+        rules = suricata.update.rule.parse_fileobj(fileobj)
+        self.assertEqual(0, len(rules))
 
     def test_parse_file_with_unicode(self):
         rules = suricata.update.rule.parse_file("./tests/rule-with-unicode.rules")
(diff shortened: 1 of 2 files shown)


tests/test_rule.py
```diff
@@ -81,15 +81,25 @@ def test_parse_fileobj(self):
...
+        rule_buf = ("""alert ( msg:"DECODE_NOT_IPV4_DGRAM"; sid:1; gid:116; rev:1; metadata:rule-type decode; classtype:protocol-command-decode;) """
+                    """alert ( msg:"DECODE_NOT_IPV4_DGRAM"; sid:1; gid:116; rev:1; metadata:rule-type decode; classtype:protocol-command-decode;)""")
         tmp = tempfile.NamedTemporaryFile()
         for i in range(2):
             tmp.write(("%s\n" % rule_buf).encode())
         tmp.flush()
         rules = suricata.update.rule.parse_file(tmp.name)
         self.assertEqual(2, len(rules))
 
+    def test_parse_rule_with_sid_error(self):
```
shivan1b: I'm a llittle skeptic about a new test function for this since this is very much redundant given we have `test_parse_fileobj`. @jasonish what do you think?

jasonish: I think I'd rather see the exception thrown from the parse() function.  Then already have the exception handler there to log it.  They way it is in this PR it looks like a bad sid will abort the parse_fileobj() which is not we want.  We want it to continue parsing the rest of the rules in the file.

So we'd want to test that parse() throws an exception if the sid is bad (I'd also handle the case where its not an integer, say a string).  Then we want to test that parse_fileobj() still parses rules after a bad rule.

vrindaaa: If my file object is say
```
alert ( msg:"DECODE_NOT_IPV4_DGRAM"; sid:3; gid:116; rev:1; metadata:rule-type decode; classtype:protocol-command-decode;)alert ( msg:"DECODE_NOT_IPV4_DGRAM"; sid:2; gid:116; rev:1; metadata:rule-type decode; classtype:protocol-command-decode;)
alert ( msg:"DECODE_NOT_IPV4_DGRAM"; sid:3; gid:116; rev:1; metadata:rule-type decode; classtype:protocol-command-decode;)alert ( msg:"DECODE_NOT_IPV4_DGRAM"; sid:2; gid:116; rev:1; metadata:rule-type decode; classtype:protocol-command-decode;)
```
why does suricata.update.rule.parse_fileobj() give the list of rules as 
`[{'enabled': True, 'action': 'alert', 'proto': None, 'source_addr': None, 'source_port': None, 'direction': None, 'dest_addr': None, 'dest_port': None, 'group': None, 'gid': 116, 'sid': 2, 'rev': 1, 'msg': 'DECODE_NOT_IPV4_DGRAM', 'flowbits': [], 'metadata': ['rule-type decode', 'rule-type decode'], 'references': [], 'classtype': 'protocol-command-decode', 'priority': 0, 'noalert': False, 'raw': 'alert ( msg:"DECODE_NOT_IPV4_DGRAM"; sid:3; gid:116; rev:1; metadata:rule-type decode; classtype:protocol-command-decode;)alert ( msg:"DECODE_NOT_IPV4_DGRAM"; sid:2; gid:116; rev:1; metadata:rule-type decode; classtype:protocol-command-decode;)', 'header': 'alert', ')alert ( msg': '"DECODE_NOT_IPV4_DGRAM"'}, {'enabled': True, 'action': 'alert', 'proto': None, 'source_addr': None, 'source_port': None, 'direction': None, 'dest_addr': None, 'dest_port': None, 'group': None, 'gid': 116, 'sid': 2, 'rev': 1, 'msg': 'DECODE_NOT_IPV4_DGRAM', 'flowbits': [], 'metadata': ['rule-type decode', 'rule-type decode'], 'references': [], 'classtype': 'protocol-command-decode', 'priority': 0, 'noalert': False, 'raw': 'alert ( msg:"DECODE_NOT_IPV4_DGRAM"; sid:3; gid:116; rev:1; metadata:rule-type decode; classtype:protocol-command-decode;)alert ( msg:"DECODE_NOT_IPV4_DGRAM"; sid:2; gid:116; rev:1; metadata:rule-type decode; classtype:protocol-command-decode;)', 'header': 'alert', ')alert ( msg': '"DECODE_NOT_IPV4_DGRAM"'}]
`
Because here isn't it essentially the same two rules repeated twice?

shivan1b: Re rules repeated: Its taken care of while building the rulemap. This parsing step is to just parse the rules into a more accessible key-value pair routine.

Re rules with sid 2 and 3: Yes. Every rule has a unique ID- (gid,sid).

vrindaaa: What does raw exactly mean in a rule? Here I think the rule with sid 3 is a part of rules['raw'] because when I'm printing rules[0]['raw'] I'm getting 
`alert ( msg:"DECODE_NOT_IPV4_DGRAM"; sid:3; gid:116; rev:1; metadata:rule-type decode; classtype:protocol-command-decode;)alert ( msg:"DECODE_NOT_IPV4_DGRAM"; sid:2; gid:116; rev:1; metadata:rule-type decode; classtype:protocol-command-decode;)`
So I don't think the rule with sid 3 is being parsed separately here?

shivan1b: raw means the rule as-is. After parsing different parts into key value pairs you get the dictionary structure like above but we preserve what the rule actually looked like.

I haven't tested it but seems like the incorrect parsing is because you're trying to write down two rules in the same line without a separator. That makes it one string.

vrindaaa: How would I add a separator between the two?

shivan1b: @vrindaaa Add `\n` to separate them as they are in any rule files. See how the rules are parsed in the `test_parse_fileobj` function.

vrindaaa: A lot of other test cases such as `test_disable_rule` and `test_parse_rule_comments_and_spaces` are failing because the rule being parsed there does not have a sid. The rule being parsed is
`# alert tcp $HOME_NET any -> $EXTERNAL_NET any (msg:"some message";)`
So should I make changes there too?

shivan1b: Yes, please.

vrindaaa: When throwing the exception from parse_fileobj(), the rest of the rules in the file are still parsed. Should I still throw the exception from parse().

PR #131, opened 2019-03-28

vrindaaa:

contribution agreement at
changes made (if applicable)

### DIFF
--- a/suricata/update/rule.py
+++ b/suricata/update/rule.py
@@ -159,6 +159,9 @@ def find_opt_end(options):
         else:
             return offset + i
 
+class BadSidError(Exception):
+    """Raises exception when sid is of type null"""
+
 def parse(buf, group=None):
     """ Parse a single rule for a string buffer.
 
@@ -281,6 +284,13 @@ def parse(buf, group=None):
 
     if rule["msg"] is None:
         rule["msg"] = ""
+
+    try:
+        if not rule["sid"]:
+            raise BadSidError("Sid cannot be of type null")
+    except BadSidError as err:
+        logger.error("Failed to parse rule: %s: %s", buf.rstrip(), err)
+        return
 
     rule["raw"] = m.group("raw").strip()
 
--- a/tests/test_rule.py
+++ b/tests/test_rule.py
@@ -45,34 +45,35 @@ class RuleTestCase(unittest.TestCase):
         self.assertEqual(rule.classtype, "trojan-activity")
 
     def test_disable_rule(self):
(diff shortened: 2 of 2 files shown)


tests/test_rule.py
```diff
@@ -45,34 +45,35 @@ def test_parse1(self):
...
         rule = suricata.update.rule.parse(rule_buf)
         self.assertFalse(rule.enabled)
         rule.enabled = True
-        self.assertEqual(str(rule), """alert tcp $HOME_NET any -> $EXTERNAL_NET any (msg:"some message";)""")
+        self.assertEqual(str(rule), """alert tcp $HOME_NET any -> $EXTERNAL_NET any (msg:"some message"; sid:1;)""")
 
     def test_parse_fileobj(self):
-        rule_buf = ("""# alert tcp $HOME_NET any -> $EXTERNAL_NET any """
-                    """(msg:"some message";)""")
+        rule_buf = ("""alert ( msg:"DECODE_NOT_IPV4_DGRAM" sid:3; gid:116; rev:1; metadata:rule-type decode; classtype:protocol-command-decode;) \n"""
```
shivan1b: There is a trailing whitespace here that needs to ne fixed. Please configure your git and your editor to not let it pass. Rest looks fine on my end. Over to @jasonish for the final review.

vrindaaa:
@jasonish can you please review this?

shivan1b:
@vrindaaa you should not force push to your branch in an open PR. Please create a new one for any changes requested. Wait for Jason's comments on this one though.

tests/test_rule.py
shivan1b: Also, this is way over the char limit of 79 per line.

suricata/update/rule.py
```diff
@@ -282,6 +285,13 @@ def parse(buf, group=None):
...
         rule["msg"] = ""
 
+    try:
+        if not rule["sid"]:
+            raise BadSidError("Sid cannot be of type null")
+    except BadSidError as err:
+        logger.error("Failed to parse rule: %s: %s", buf.rstrip(), err)
+        return
+
     rule["raw"] = m.group("raw").strip()
```
jasonish: I think the above should just be:
```
    if not rule["sid"]:
        raise BadSidError("Sid cannot be of type null")
```

`parse_fileobj` will take care logging the error and continuing on.  I think its important that `parse()` raise this error to the caller as `None` implies the provided input didn't look like a rule.  But if it did look like a rule and its was missing a sid, we should raise an error.

Optional: check that the sid can also be parsed as an integer and raise the exception if it can't.

jasonish:
We should also have a test to check that the exception is raised when a rule is parsed without a sid.

PR #142, opened 2019-04-09
make sure that noalert is set in newly enabled rules v4

King-Konsto:
This commit adds functionality that ensures that previously
disabled rules enabled by flowbit dependencies will receive
the noalert option.

-- thank you.

  contribution agreement at
  changes made (if applicable)

Link
to
https://redmine.openinfosecfoundation.org/issues/2906

Describe changes:
- This commit adds functionality that ensures that previously disabled rules enabled by flowbit dependencies will receive the flowbits:noalert option.
- Set this to default behavior after the redmine ticket discussion.
- follow up to https://github.com/OISF/suricata-update/pull/141

### DIFF
--- a/suricata/update/rule.py
+++ b/suricata/update/rule.py
@@ -146,6 +146,8 @@ class Rule(dict):
         return self.format()
 
     def format(self):
+        if self.noalert and not "noalert;" in self.raw:
+            self.raw = re.sub(r'( *sid\: *[0-9]+\;)', r' noalert;\1', self.raw)
         return u"%s%s" % (u"" if self.enabled else u"# ", self.raw)
 
 def find_opt_end(options):
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -725,6 +725,7 @@ def resolve_flowbits(rulemap, disabled_rules):
                     "Enabling previously disabled rule for flowbits: %s" % (
                         rule.brief()))
             rule.enabled = True
+            rule.noalert = True
             flowbit_enabled.add(rule)
     logger.info("Enabled %d rules for flowbit dependencies." % (
         len(flowbit_enabled)))


suricata/update/rule.py
```diff
@@ -146,6 +146,8 @@ def __str__(self):
         return self.format()
 
     def format(self):
+        if self.noalert and not "noalert;" in self.raw:
+            self.raw = re.sub(r'( *sid\: *[0-9]+\;)', r' noalert;\1', self.raw)
```
victorjulien: does this mean a space between 'sid:' and the value is required?

King-Konsto: No, it's not required. It's just optional so that the regex doesn't fail if someone wrote a rule with a space between "sid:" and the number, eg. "sid: 1234567;".

victorjulien: Ok thanks. @shivan1b do we have a way to add an automated test?

shivan1b: Yes. We can add a test method to `test_rule.py` satisfying the conditions above in the `format()` method. We could also modify existing test `test_toggle_rule` with a new usecase for `noalert` set to True.
@King-Konsto do you mind adding a test? Let me know if you need any assistance with that.

King-Konsto: Yes, I can do that. Can you guide me to an example of an existing test and explain what you think I should do please?

jasonish: One template test could be `def test_disable_rule(self)`.  Parse a rule, set the noalert on it, reformat it, and make sure it parses again?

shivan1b: Also, @King-Konsto it would be great if you could add a test for `resolve_flowbits` function since that is the function where you actually made a change. In order to write the test, try to replicate the function signature (here `rulemap` and `disabled_rules`). These should be hardcoded by you so that tests do not give different results in different settings.
You can take reference from any test https://github.com/OISF/suricata-update/blob/master/tests/test_rule.py to understand what your test should look like. Let me know if you need any more assistance.

King-Konsto:
Close and open new PR: https://github.com/OISF/suricata-update/pull/147

suricata/update/rule.py
King-Konsto: Thanks for your help. I have added test cases. Please have a look at https://github.com/OISF/suricata-update/pull/147

PR #150, opened 2019-05-13
Add "show-defaults" command

shivan1b:
Add a command "show-defaults" that shows current default configuration
for suricata-update.

**Sample**

```
└─ $ ▶ ./bin/suricata-update --show-defaults
13/5/2019 -- 16:47:48 - <Info> -- Loading /etc/suricata/update.yaml
13/5/2019 -- 16:47:48 - <Info> -- Using data-directory /var/lib/suricata.
13/5/2019 -- 16:47:48 - <Info> -- Using Suricata configuration /etc/suricata/suricata.yaml
13/5/2019 -- 16:47:48 - <Info> -- Using /etc/suricata/rules for Suricata provided rules.
disable-conf                  /etc/suricata/disable.conf
enable-conf                   /etc/suricata/enable.conf
drop-conf                     /etc/suricata/drop.conf
modify-conf                   /etc/suricata/modify.conf
sources                       Not set
local                         Not set
ignore
                              *deleted.rules
subcommand                    update
show-defaults                 1
```
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2938

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -32,6 +32,7 @@ import shutil
 import glob
 import io
 import tempfile
+from collections import Iterable
 
 try:
     # Python 3.
@@ -1018,6 +1019,20 @@ def check_output_directory(output_dir):
                 "Failed to create directory %s: %s" % (
                     output_dir, err))
 
+
+def show_defaults():
+    row_format = "{:<30}" * 2
+    for conf_key, conf_val in config._config.items():
+        if conf_val and isinstance(conf_val, Iterable) and \
+                not isinstance(conf_val, str):
+            print(conf_key)
+            if isinstance(conf_val, list):
+                for i in conf_val:
+                    print("{:>30}{}".format("", i))
+        else:
+            print(row_format.format(conf_key, conf_val if conf_val else "Not set"))
+
+
 def _main():
     global args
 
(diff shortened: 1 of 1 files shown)


suricata/update/main.py
```diff
@@ -1018,6 +1019,20 @@ def check_output_directory(output_dir):
...
+    row_format = "{:<30}" * 2
+    for conf_key, conf_val in config._config.items():
+        if conf_val and isinstance(conf_val, Iterable) and \
+                not isinstance(conf_val, str):
+            print(conf_key)
+            if isinstance(conf_val, list):
+                for i in conf_val:
+                    print("{:>30}{}".format("", i))
+        else:
+            print(row_format.format(conf_key, conf_val if conf_val else "Not set"))
```
shivan1b: hmm this should not have the internal information like "subcommand". Needs a spin.

shivan1b: On second thoughts, this is probably a redundant feature. `-v` gives most of the required information and now we even have an `--offline` option in case we do not want to update the rules while getting this info.

jasonish: I could live without this feature.  But its not a bad feature.  Some command with the following output could be useful:
```
Using Suricata: /usr/bin/suricata
Using Suricata configuration file: /etc/suricata/suricata.yaml
Config files Suricata-update would load...
Writing rules to ....
```
Thats pretty rough, but gets the idea across.

Another idea would be a `--dry-run` that goes through all the motions, but doesn't actually do anything. Could be tricky though.

inashivb: Should the `--dry-run` feature be implemented instead then?

inashivb:
Closing as stale.

PR #152, opened 2019-05-17
Improve permission warning

vagishagupta23:
Improve permission warning when Suricata-update runs with the wrong user

When suricata-update runs with a non-root user, it gives an ugly traceback.
To avoid those ugly tracebacks, the permission of the configuration file
and rule file is checked where the files are loaded from configuration
file directory and rule output directory respectively and exit cleanly
with an error in the log.

-- thank you.

  contribution agreement at
  changes made (if applicable)

Link
to
[redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2875

Describe changes:
-
-
-

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -1280,6 +1280,28 @@ def _main():
         logger.info("Loading %s.", drop_conf_filename)
         drop_filters += load_drop_filters(drop_conf_filename)
 
+    # Check permission for configuration file.
+    if config.get("suricata-conf") and \
+       os.path.exists(config.get("suricata-conf")):
+        try:
+            with open(config.get("suricata-conf")) as fileobj:
+                pass
+        except IOError as err:
+            logger.error(
+                   "permission denied for: %s", config.get("suricata-conf"))
+            return 1
+
+    # Check permission for output rule file.
+    if os.path.exists(config.get_output_dir()):
+        output_filename = os.path.join(
+                config.get_output_dir(), DEFAULT_OUTPUT_RULE_FILENAME)
+        try:
+            with open(output_filename) as fileobj:
+                pass
+        except IOError as err:
+            logger.error("permission denied for: %s", output_filename)
+            return 1
+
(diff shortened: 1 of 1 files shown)


suricata/update/main.py
```diff
@@ -1280,6 +1280,28 @@ def _main():
...
+    if os.path.exists(config.get_output_dir()):
+        output_filename = os.path.join(
+                config.get_output_dir(), DEFAULT_OUTPUT_RULE_FILENAME)
+        try:
+            with open(output_filename) as fileobj:
+                pass
+        except IOError as err:
+            logger.error("permission denied for: %s", output_filename)
+            return 1
+
```
shivan1b: Hint: Read the traceback for redmine bug and the traceback you're getting carefully. Feedback:
1. 22 lines of code just to check file permissions of two different files/directories - Probably not such great idea.
2. Both of these `if` blocks are exactly the same except for the conditions and `get` param. - This means too much of code repetition. Python gives you the ability to create dynamic var/func names. That could be used. And mix it up with a for loop and save approx. 10 lines of code here.
3. I'm sorry for having been unclear in the last review.
Wrap the call to file or dir in a try except block. - This means that put the try except block around the operations that are being performed on them *wherever* they are being performed. Randomly putting `open` checks with a `pass` statement is not the good way to go about it.

You probably do not require (1) and (2) anymore but keep them in mind for future work maybe.
Could you please look up the exact places where exceptions are coming from and make appropriate changes there?
Any questions, please let me know. Thanks for your work. :)

vagishagupta23: @shivan1b Thanks for the feedback.
I think I figured out my mistake thanks. I am little confused here that should I exit everywhere I am putting try except block wherever the operations are being performed?
Also, should there be any message that I need to log with the error?

vagishagupta23:
Created a new PR #153

PR #153, opened 2019-05-20
Improve permission warning

vagishagupta23:
Improve permission warning when Suricata-update runs with the wrong user

When suricata-update runs with a non-root user, it gives an ugly traceback.
To avoid those ugly tracebacks, the permission of the file is checked earlier
and try except block is put around the operations that are being performed on
the file and exit cleanly with an error in the log.

-- thank you.

  contribution agreement at
  changes made (if applicable)

Link
to
[redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:https://redmine.openinfosecfoundation.org/issues/2875

Describe changes:
-
-
-

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -353,6 +353,11 @@ class Fetch:
         net_arg = url
         url = url[0] if isinstance(url, tuple) else url
         tmp_filename = self.get_tmp_filename(url)
+        try:
+            with open(tmp_filename, 'wb') as fileobj:
+                pass
+        except IOError as err:
+            logger.warning("Failed to open %s: %s" % (tmp_filename, err))
         if not config.args().force and os.path.exists(tmp_filename):
             if not config.args().now and \
                time.time() - os.stat(tmp_filename).st_mtime < (60 * 15):
@@ -381,6 +386,9 @@ class Fetch:
                     "will use latest cached version: %s", url, err)
                 return self.extract_files(tmp_filename)
             raise err
+        except IOError as err:
+            logger.error("Failed to copy file %s", err)
+            sys.exit(1)
         except Exception as err:
             raise err
         self.progress_hook_finish()


suricata/update/main.py
```diff
@@ -353,6 +353,11 @@ def fetch(self, url):
         net_arg = url
         url = url[0] if isinstance(url, tuple) else url
         tmp_filename = self.get_tmp_filename(url)
+        try:
+            with open(tmp_filename, 'wb') as fileobj:
```
shivan1b: 1. The file was not being created here so this definitely cannot be the place where the error was coming from. Here we are just creating a temporary **file name**.
2.  The error comes on the step where we try to `shutil.copyfile(tmp_fileobj.name, tmp_filename)` i.e. actually opening the file to write the content into it. Read about it here: https://docs.python.org/3/library/shutil.html#shutil.copyfile

suricata/update/main.py
```diff
@@ -381,6 +386,9 @@ def fetch(self, url):
                     "will use latest cached version: %s", url, err)
                 return self.extract_files(tmp_filename)
             raise err
+        except IOError as err:
+            logger.error("Failed to copy file %s", err)
+            sys.exit(1)
```
shivan1b: So, this block would have sufficed. However,
`err` will never have the filename. So, the error output would look absurd. Something like,
```
21/5/2019 -- 18:19:05 - <Error> -- Failed to copy file [Errno 13] Permission denied: u'/var/lib/suricata/update/cache/92a59cfd53a431780ba3036effa26f9b-sslblacklist.rules'
```
This should be changed.

vagishagupta23:
> @vagishagupta23 good improvements. :) Does this also take care of the permission error you were getting for `suricata.yaml`?

@shivan1b Thanks for the feedback :) 
No, this PR doesn't take care of the suricata.yaml permission error.

suricata/update/main.py
vagishagupta23: Yes, I was thinking to explicitly check for the write permissions of dest location in shutil.copyfile. After going through the link, I understood that it itself opens the file and checks that the destination location must be writable. Learned a new thing today. Thanks :)

suricata/update/main.py
vagishagupta23: Sorry, but I didn't get this. Do you mean that error message should have the filename where it is being copied?

shivan1b: Yes. According to the string that is being written in the error message. i.e. `Failed to copy file ...` one would be expecting a file name following this line and not the error message.

shivan1b:
@vagishagupta23 thanks for informing. Could you please add that check in the same PR?

vagishagupta23:
> @vagishagupta23 thanks for informing. Could you please add that check in the same PR?

Sure.

vagishagupta23:
Created a new PR #154.

PR #157, opened 2019-05-28
Add check for output dir for rules

shivan1b:
This patch assumes that any dir which has a `*.rules` file or is an empty
dir can only be the valid choices for an output rule directory.

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2691

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -19,6 +19,7 @@ from __future__ import print_function
 
 import sys
 import re
+import errno
 import os.path
 import logging
 import argparse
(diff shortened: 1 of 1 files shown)


suricata/update/main.py
```diff
@@ -1390,6 +1399,11 @@ def _main():
     # Backup the output directory.
     logger.info("Backing up current rules.")
     backup_directory = util.mktempdir()
+    if not (any(fname.endswith('.rules') for fname in os.listdir(
```
shivan1b: @jasonish do you think it makes sense to consider a non-empty directory with does not have any `*.rules` file a not ideal directory for writing rules to?
Code can be improved, it'll come in the next PR after having solved conflicts if this makes sense to you.

jasonish: I think this will get tricky to maintain.  For example, at some point we'll have to move support files into the rules directory (https://redmine.openinfosecfoundation.org/issues/2688). These support files could potential have any possible filename. So that also means we may need to do conflict resolution there as well.

Adding some disucssion to:
https://redmine.openinfosecfoundation.org/issues/2691

PR #162, opened 2019-06-04
Log a warning on duplicate SID

vagishagupta23:
Currently when suricata-update encounters a rules with duplicate SIDs,
it silently uses the one with the higher revision without logging the
warnings.
On duplicate SID, warnings are logged for equal as well as different
revisions.

-- thank you.

  contribution agreement at
  changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2879

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -692,7 +692,14 @@ def build_rule_map(rules):
         if rule.id not in rulemap:
             rulemap[rule.id] = rule
         else:
+            if rule["rev"] == rulemap[rule.id]["rev"]:
+                logger.warning(
+                    "Found duplicate rule SID {}, keeping existing rule."
+                    .format(rule.sid))
             if rule["rev"] > rulemap[rule.id]["rev"]:
+                logger.warning(
+                    "Found duplicate rule SID {}, "
+                    "keeping one with higher revision.".format(rule.sid))
                 rulemap[rule.id] = rule
 
     return rulemap


suricata/update/main.py
```diff
@@ -692,7 +692,14 @@ def build_rule_map(rules):
...
             rulemap[rule.id] = rule
         else:
+            if rule["rev"] == rulemap[rule.id]["rev"]:
+                logger.warning(
+                    "Found duplicate rule SID {}, keeping existing rule."
+                    .format(rule.sid))
             if rule["rev"] > rulemap[rule.id]["rev"]:
+                logger.warning(
+                    "Found duplicate rule SID {}, "
+                    "keeping one with higher revision.".format(rule.sid))
```
victorjulien: should this be something like 'keeping *the* one with the *highest* revision." ?

victorjulien: Maybe instead of ```the one``` it reads better to have ```the rule``` ?

jasonish: In the first case, I think I'd like a little more information:
```
Found duplicate rule SID {} with same revision, keeping first rule seen.
```

And like Victor said for the second case:
```
Found duplicate rule SID {}, keeping rule with greater revision."
```

vagishagupta23: Thanks. This sounds much better. Addressing the changes in a new PR.

vagishagupta23:
A new PR is here #183.

Kolano:
> Do not use the path ``/etc/suricata`` and ``/var/lib`` for example, since it could be any other path that user defines during configuration and is by default ``/usr/local``. Please check your configuration directory for the path and use that accordingly.

The wording of that isn't very good. I'd suggest...
> The paths ``/etc/suricata`` and ``/var/lib`` above are used in the default configuration and are dependent on paths set during compilation / the suricata.yaml and update.yaml files, By default these paths are set to ``/usr/local``. Please check your configuration for appropriate paths.

shivan1b:
@Kolano wrong issue for the comment maybe?

Kolano:
Sorry, yeah, meant for #168

PR #165, opened 2019-06-10
Redo variable and function names reserved for Python

vagishagupta23:
Changed the conflicting variable name `filter` and function name
`filter()` which suricata-update uses as it is reserved for use in 
Python standard modules.

-- thank you.

  contribution agreement at
  changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3012

Describe changes:
- Changed variable and function names `filter` in `main.py`
- Changed variable and function names `filter` in `test_main.py`

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -252,7 +252,7 @@ class ModifyRuleFilter(object):
     def match(self, rule):
         return self.matcher.match(rule)
 
-    def filter(self, rule):
+    def modify_filter(self, rule):
         modified_rule = self.pattern.sub(self.repl, rule.format())
         parsed = suricata.update.rule.parse(modified_rule, rule.group)
         if parsed is None:
@@ -291,7 +291,7 @@ class DropRuleFilter(object):
             return False
         return self.matcher.match(rule)
 
-    def filter(self, rule):
+    def drop_filter(self, rule):
         drop_rule = suricata.update.rule.parse(re.sub("^\w+", "drop", rule.raw))
         drop_rule.enabled = rule.enabled
         return drop_rule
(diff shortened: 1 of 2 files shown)


suricata/update/main.py
```diff
@@ -252,7 +252,7 @@ def __init__(self, matcher, pattern, repl):
     def match(self, rule):
         return self.matcher.match(rule)
 
-    def filter(self, rule):
+    def modify_filter(self, rule):
```
shivan1b: Repeating the functionality of a class in its methods is usually a bad practice. What I mean is, when you want to access this function, you'd probably do `ModifyRuleFilter.modify_filter()` see the `modify` twice? Was it even needed? Probably not. Because in this case `filter` has been defined under a separate namespace. Hence, `ModifyRuleFilter.filter()` makes a better candidate. However, since `filter()` is reserved, we should in any circumstances avoid using it directly. The next best thing would be `_filter()`. But, it is used to indicate private use methods.
Conclusion: I think something should be changed here, I can't think of a good function name. Please find something?
Also, if you think otherwise or find some guide/docs that provide a clear picture of naming scheme in such cases, please let me know.

suricata/update/main.py
```diff
@@ -291,7 +291,7 @@ def match(self, rule):
             return False
         return self.matcher.match(rule)
 
-    def filter(self, rule):
+    def drop_filter(self, rule):
```
shivan1b: Same here.

suricata/update/main.py
vagishagupta23: Hey @shivan1b I tried going through many guide/docs for such cases but couldn't find one. I could come up with a few names for the `filter()` function like `filterRule()`, `drop()/modify()`, `dropRule()/modifyRule()`and `filtered()`. Do you think any of these could be used?

shivan1b: Hmm I guess then we should leave this as-is. Let @jasonish decide if this looks OK.

jasonish: So part of the idea is that a filter follows an interface, so executing doesn't require knowledge of what its doing.  I'd prefer to rename the method to something more generic..  `run()`, or `execute()` seem to be common names in other languages where filters are part of a processing pipeline.

vagishagupta23: Thanks for clarifying @jasonish. Will fix it in a new PR.

vagishagupta23:
Created a new PR #174.

PR #166, opened 2019-06-12
Docs: Setup directories with correct permissions

vagishagupta23:
Documentation updated with the steps to setup the above directories
with the correct permissions.

-- thank you.

  contribution agreement at
  changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3038

### DIFF
--- a/doc/quickstart.rst
+++ b/doc/quickstart.rst
@@ -52,8 +52,37 @@ One option is to simply run ``suricata-update`` as root or with
           the ``suricata`` group then add users to the ``suricata``
           group.
 
-	  More documentation will be provided about this, including a
-	  tool to verify and maybe setup the permissions.
+Steps to setup the above directories with the correct permissions:
+
+First, create a group ``suricata``::
+
+    sudo groupadd suricata
+
+Next, change the group of the directories and its files recursively::
+
+    sudo chgrp -R suricata /etc/suricata
+    sudo chgrp -R suricata /var/lib/suricata/rules
+    sudo chgrp -R suricata /var/lib/suricata/update
+
+Setup the directories with the correct permissions for the ``suricata``
+group::
+
+    sudo chmod -R g+r /etc/suricata/
+    sudo chmod -R g+rw /var/lib/suricata/rules
+    sudo chmod -R g+rw /var/lib/suricata/update
+
+Now, add user to the group::
+
+    sudo usermod -a -G suricata username
+
+Verify whether group has been changed::
+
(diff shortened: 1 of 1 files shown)


doc/quickstart.rst
```diff
@@ -52,8 +52,37 @@ One option is to simply run ``suricata-update`` as root or with
...
+
+    sudo usermod -a -G suricata username
+
+Verify whether group has been changed::
+
+    ls -al /etc/suricata
+    ls -al /var/lib/suricata/rules
+    ls -al /var/lib/suricata/update
+
+Reboot your system. Run ``suricata-update`` without a sudo to check
```
shivan1b: Do we need a reboot?

doc/quickstart.rst
```diff
@@ -52,8 +52,37 @@ One option is to simply run ``suricata-update`` as root or with
...
+Next, change the group of the directories and its files recursively::
+
+    sudo chgrp -R suricata /etc/suricata
+    sudo chgrp -R suricata /var/lib/suricata/rules
+    sudo chgrp -R suricata /var/lib/suricata/update
+
+Setup the directories with the correct permissions for the ``suricata``
+group::
+
+    sudo chmod -R g+r /etc/suricata/
```
shivan1b: Hmm most of the configuration like update.yaml, modify.yaml goes here so it should not just have read access.

doc/quickstart.rst
vagishagupta23: @shivan1b It doesn't run on my system without a reboot.

doc/quickstart.rst
vagishagupta23: So, it should have both read and write access? Then, there should also be a change in the docs where it is written `Directory /etc/suricata: read access`. Should I consider updating it?

vagishagupta23:
> @vagishagupta23 Minor changes requested. Apart from this, please note that we should probably not use `/etc/suricata` and `/var/lib` for example, since it could be any other path that user defines during configuration and is by default `/usr/local`. Maybe, add a note that people should check their configuration directory and use that accordingly?

Yes, I was also confused about it. Thanks for confirming. Will address these changes.

> This probably won't be a problem with experienced Suricata users but we should write the docs such that beginners do not end up making a lot of mistakes.
> 
> Looks awesome other than this. Thanks! :)

vagishagupta23:
A new PR is here #168.

PR #170, opened 2019-06-20
Add "offline" command

vagishagupta23:
Add a command line option `--offline` that uses locally cached
latest version of rules without trying to download rules from
sources.

-- thank you.

  contribution agreement at
  changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2864

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -353,6 +353,13 @@ class Fetch:
         net_arg = url
         url = url[0] if isinstance(url, tuple) else url
         tmp_filename = self.get_tmp_filename(url)
+        if config.args().offline:
+            logger.info("Proceeding offline, will use latest cached version.")
+            if not os.path.exists(tmp_filename):
+                logger.error("Can't proceed offline, "
+                             "source has not yet been downloaded.")
+                sys.exit(1)
+            return self.extract_files(tmp_filename)
         if not config.args().force and os.path.exists(tmp_filename):
             if not config.args().now and \
                time.time() - os.stat(tmp_filename).st_mtime < (60 * 15):
@@ -985,7 +992,7 @@ def load_sources(suricata_version):
     # If --etopen is on the command line, make sure its added. Or if
     # there are no URLs, default to ET/Open.
     if config.get("etopen") or not urls:
-        if not urls:
+        if not config.args().offline and not urls:
             logger.info("No sources configured, will use Emerging Threats Open")
         urls.append(sources.get_etopen_url(internal_params))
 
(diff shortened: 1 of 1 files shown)


suricata/update/main.py
```diff
@@ -985,7 +992,7 @@ def load_sources(suricata_version):
     # If --etopen is on the command line, make sure its added. Or if
     # there are no URLs, default to ET/Open.
     if config.get("etopen") or not urls:
-        if not urls:
+        if not config.args().offline and not urls:
```
shivan1b: I believe this would be taken care of in the fetch function itself? So, this check should not be required.

suricata/update/main.py
```diff
@@ -1129,6 +1136,9 @@ def _main():
     
     update_parser.add_argument("--no-merge", action="store_true", default=False,
                                help="Do not merge the rules into a single file")
+    update_parser.add_argument("--offline", action="store_true", default=False,
```
shivan1b: One more thing, I do not think you require a `default` while mentioning the `action` here, `store_true` means default is false and when specified, it shall be true. Read and try: https://docs.python.org/3/library/argparse.html#action

suricata/update/main.py
vagishagupta23: If I remove this check the output would be something like this:
```
25/6/2019 -- 13:05:32 - <Info> -- No sources configured, will use Emerging Threats Open
25/6/2019 -- 13:05:32 - <Info> -- Proceeding offline, will use latest cached version.
```
IMHO, we are using offline mode there should be no need of logging this info `No sources configured, will use Emerging Threats Open` as there is no requirement of using etopen. This was the reason adding the check here. What are your thoughts on it?

suricata/update/main.py
```diff
@@ -353,6 +353,13 @@ def fetch(self, url):
         net_arg = url
         url = url[0] if isinstance(url, tuple) else url
         tmp_filename = self.get_tmp_filename(url)
+        if config.args().offline:
+            logger.info("Proceeding offline, will use latest cached version.")
```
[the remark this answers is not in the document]
vagishagupta23: @shivan1b I'm not able to figure out why this happened. I am not getting `Proceeding offline, will use latest cached version` 3 times. Which command are you running?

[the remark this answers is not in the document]
shivan1b: @vagishagupta23 this function is called per source url. You probably have just one source enabled. Check using `suricata-update list-enabled-sources`.

suricata/update/main.py
shivan1b: yes this makes sense. Thanks for clarifying.

vagishagupta23:
Fixed changes in a new PR #172.

PR #171, opened 2019-06-21
Separate out parsers

vagishagupta23:
Parsers are separated out of the `main.py` module to a new `parsers.py`
module to make code cleaner and compact.

This patch is heavily based on Vidushi's work.
Thanks to Vidushi Agrawal <vidushi229@gmail.com>

-- thank you.

  contribution agreement at
  changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2874

Describe changes:
- Created a new module `parsers.py` with reduced repetitions of
  `add_argument` by storing the arguments in a tuple and added
  the parsers to the loop.
- Also, parsers.py is broken into smaller functions based on different
  parsers i.e. `global_parser`, `update_parsers` and `commands`.


suricata/update/parsers.py
```diff
@@ -0,0 +1,245 @@
...
+        if not hasattr(args, arg):
+            setattr(args, arg, getattr(global_args, arg))
+        elif hasattr(args, arg) and getattr(args, arg) is None:
+            setattr(args, arg, getattr(global_args, arg))
+
+    # Go verbose or quiet sooner than later.
+    if args.verbose:
+        logger.setLevel(logging.DEBUG)
+    if args.quiet:
+        logger.setLevel(logging.WARNING)
```
shivan1b: Hmm I'm not sure how is this working but from what I understand, `logger.setLevel` was for the main file's logger instance and since the level was not set there, this should not have worked. Can you figure out why did this work?

vagishagupta23: After reading about the loggers I could understand the following (Please correct me if I am wrong):
1. When loggers are instantiated using `logging.getLogger()` without specifying a name (as `logging.getLogger(name)`)  it applies the configuration to the __root logger__ (in main module) and also to the parser.py module logger due to which `logger.setLevel` (for root logger) here is configured to DEBUG and/or WARNING. So it worked. 
2. If I replace `getLogger()` in the main module by `getLogger(__name__)`, the logger in the parsers.py module will become a child of the root logger and it will thus inherit the root logger configuration and as the root logger has `setLevel` configured to `logging.INFO` it will log INFO messages.
I referred these links:
- https://docs.python.org/3/howto/logging.html#loggers
- https://docs.python.org/3/library/logging.html#logger-objects
- https://stackoverflow.com/questions/50714316/how-to-use-logging-getlogger-name-in-multiple-modules/50715155#50715155

shivan1b: Thanks for the explanation. I checked and it seems they are indeed the same root logger instances in both the files. However, since we're doing most of the logger related work in main, I'm a little inclined to setting these levels in the main itself. They are not anyway a part of parser code.

vagishagupta23:
Addressed changes in a new PR #173.

PR #180, opened 2019-07-15
Add "check-versions" command

vagishagupta23:
Add a `--check-versions` command that checks the version of suricata
and logs if the versions are up to date, outdated or EOL.

-- thank you.

  contribution agreement at
  changes made (if applicable)


suricata/update/main.py
```diff
@@ -1018,6 +1018,26 @@ def check_output_directory(output_dir):
...
+    if not os.path.exists(index_filename):
+        logger.warning("No index exists, will use bundled index.")
+        logger.warning("Please run suricata-update update-sources.")
+    index = sources.Index(index_filename)
+    version = index.get_versions()
+    if suricata_version.full in version['suricata']['recommended'] or \
+            "dev" in str(suricata_version.full):
+        logger.info("Suricata version %s is up to date", suricata_version.full)
+    elif suricata_version.short in version['suricata'] and \
+            suricata_version.full not in version['suricata']['4.1']:
```
victorjulien: I'm surprised to see a hardcoded version here. Shouldn't this be fully handled by the index file?

We know the running Suricata version, so we can take '4.1.4', reduce it to '4.1', look up the '4.1' entry in the index and check if our 4.1.4 matches the version we have for '4.1'.

vagishagupta23: Thanks for catching this. Going to fix this issue.

vagishagupta23: Could you please suggest what should be done in case if the Suricata version is greater than the recommended version. Like if someone uses the following command:
`sudo ./bin/suricata-update --check-versions --suricata-version 5.2.3`

shivan1b: @vagishagupta23 I think the recommended version will always be the latest stable (in our intel index and that is kind of our responsibility to manage). So unless its dev or pre-release, this situation should not arise.
For example, currently, the stable release and the recommended version is `4.1.4` (check here: https://github.com/OISF/suricata/releases/tag/suricata-4.1.4) however there can be some pre-releases like there is `5.0.0-beta1` now but this is not the recommended version because its not stable.

I think it would be helpful for you to have a look at the previous releases and understand the difference and naming schemes: https://github.com/OISF/suricata/releases

If required, we can always change our index to accomodate some more information. ;)

suricata/update/main.py
```diff
@@ -1018,6 +1018,26 @@ def check_output_directory(output_dir):
...
+        logger.warning("Suricata version %s is outdated. "
+                       "Please upgrade to %s.", suricata_version.full,
+                       version['suricata']['recommended'])
+    else:
+        logger.warning("Suricata version %s is End of life. "
```
shivan1b: nit: `s/is End of life/has reached EOL`

suricata/update/main.py
jasonish: > This makes me realize that `suricata-version` functionality is probably not working fine? No matter what version I provide, it says in the logs "Forcing Suricata version to 5.6". 5.6 does not even exist. Seems like it just gives the output without checking if that Suricata version exists on the system. @jasonish is this the expected behavior?

More or less.  This is to force a specific version for the case where you are using Suricata-Update without having Suricata installed, or building a ruleset for a different version of Suricata that you have installed.

One use case is running Suricata-Update on one system, then using some deployment/provisioning system like Ansible to push the rules out to your Suricata installations.

shivan1b: OK. Thanks! :)

vagishagupta23: > @vagishagupta23 I think the recommended version will always be the latest stable (in our intel index and that is kind of our responsibility to manage). So unless its dev or pre-release, this situation should not arise.
> For example, currently, the stable release and the recommended version is `4.1.4` (check here: https://github.com/OISF/suricata/releases/tag/suricata-4.1.4) however there can be some pre-releases like there is `5.0.0-beta1` now but this is not the recommended version because its not stable.
> 
> I think it would be helpful for you to have a look at the previous releases and understand the difference and naming schemes: https://github.com/OISF/suricata/releases
> 
> If required, we can always change our index to accomodate some more information. ;)

Thanks. These links would be helpful to understand the naming schemes. :)

suricata/update/data/index.py
```diff
@@ -1 +1 @@
...
...Positive Technologies Attack Detection Team ruleset'}, 'sslbl/ssl-fp-blacklist': {'url': 'https://sslbl.abuse.ch/blacklist/sslblacklist.rules', 'vendor': 'Abuse.ch', 'description': 'The SSL Blacklist (SSLBL) is a project of abuse.ch with the goal of detecting malicious SSL connections, by identifying and blacklisting SSL certificates used by botnet C&C servers. In addition, SSLBL identifies JA3 fingerprints that helps you to detect & block malware botnet C&C communication on the TCP layer.\n', 'license': 'Non-Commercial', 'summary': 'Abuse.ch SSL Blacklist'}, 'tgreen/hunting': {'vendor': 'tgreen', 'description': 'Heuristic ruleset for hunting. Focus on anomaly detection and showcasing latest engine features, not performance.\n', 'license': 'GPLv3', 'url': 'https://raw.githubusercontent.com/travisbgreen/hunting-rules/master/hunting.rules', 'min-version': '4.1.0', 'summary': 'Threat hunting rules'}, 'etnetera/aggressive': {'url': 'https://security.etnetera.cz/feeds/etn_aggressive.rules', 'vendor': 'Etnetera a.s.', 'min-version': '4.0.0', 'license': 'MIT', 'summary': 'Etnetera aggressive IP blacklist'}}, 'version': 1, 'versions': {'suricata': {'recommended': '4.1.4', '4.1': '4.1.4'}}}
```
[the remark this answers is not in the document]
vagishagupta23: Yes. I would add it in a follow-up PR. 
Should I also check versions for the suricata-update in the code?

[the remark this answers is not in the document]
jasonish: > Should I also check versions for the suricata-update in the code?

I'd do that as another feature, after this was is all worked out.

vagishagupta23:
Replaced by #182.

PR #186, opened 2019-07-23
Add "check-versions" subcommand

vagishagupta23:
Add a `suricata-update check-versions` subcommand that checks
the version of suricata and logs if the versions are up to date, 
outdated or EOL.

**Sample**

```
└─ $ ▶ ./bin/suricata-update check-versions
23/7/2019 -- 17:55:42 - <Info> -- Using data-directory /usr/local/var/lib/suricata.
23/7/2019 -- 17:55:42 - <Info> -- Using Suricata configuration /usr/local/etc/suricata/suricata.yaml
23/7/2019 -- 17:55:42 - <Info> -- Using /usr/local/etc/suricata/rules for Suricata provided rules.
23/7/2019 -- 17:55:42 - <Info> -- Found Suricata version 5.0.0-dev at /usr/local/bin/suricata.
23/7/2019 -- 17:55:42 - <Info> -- Suricata version 5.0.0-dev is up to date
```

-- thank you.

  contribution agreement at
  changes made (if applicable)


suricata/update/commands/checkversions.py
```diff
@@ -0,0 +1,57 @@
...
+
+def check_version(suricata_version):
+    index_filename = sources.get_index_filename()
+    if not os.path.exists(index_filename):
+        logger.warning("No index exists, will use bundled index.")
+        logger.warning("Please run suricata-update update-sources.")
+    index = sources.Index(index_filename)
+    version = index.get_versions()
+    recommended = engine.parse_version(version['suricata']['recommended'])
+    if suricata_version.full in recommended.full:
```
victorjulien: is this an exact match? or would '4.1.4' in '4.1.4.1' also match? We don't use the latter notation at this time, but if we can have the logic be robust against such changes I think it would be great.

jasonish: This uses the existing version parser where `.full` is the just the raw version string passed in, provided it parsed OK.  Right now this only supports 4.1.4, not 4.1.4.1.  I'd suggest doing that as a separate task.

victorjulien: Not sure I follow. An exact string match would accomplish both, right?

victorjulien: It would also make sure '3.1' in '4.3.1' wouldn't match.

vagishagupta23: > is this an exact match? or would '4.1.4' in '4.1.4.1' also match? We don't use the latter notation at this time, but if we can have the logic be robust against such changes I think it would be great.

It's not the exact match. Here '4.1.4' in '4.1.4.1' would match if we use '4.1.4.1' as recommended version. I think we should have an exact string match here.

jasonish:
Example failure.  Given an index.yaml containing:
```
versions:
  suricata:
    recommended: 5.0.0
    "5.0": 5.0.0
    "4.1": 4.1.4
```
And I'm running 4.1.4, I get the message:
```
Suricata version 4.1.4 has reached EOL. Please upgrade to 5.0.0.
```
However, 4.1.4 is not EOL and is still a valid version.  I should get a message about my version of Suricata being outdated instead.

Had I been running 4.0.x, the EOL message is correct.

suricata/update/commands/checkversions.py
```diff
@@ -0,0 +1,57 @@
+# Copyright (C) 2017 Open Information Security Foundation
```
jasonish: 2019

vagishagupta23:
Changes addressed in #195.

PR #191, opened 2019-08-20
Add no-checksum option

vagishagupta23:
Added a --no-checksum option to the add-source command and an optional
argument "no_checksum" in SourceConfiguration class for add-sources and
enabled-sources.
Also, a check is added to skip downloading the checksum URL if the
source is configured  with no-checksum true.

-- thank you.

  contribution agreement at
  changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3100

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -352,6 +352,7 @@ class Fetch:
 
     def fetch(self, url):
         net_arg = url
+        no_checksum = url[2] if isinstance(url, tuple) else False
         url = url[0] if isinstance(url, tuple) else url
         tmp_filename = self.get_tmp_filename(url)
         if config.args().offline:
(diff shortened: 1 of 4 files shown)


suricata/update/main.py
```diff
@@ -352,6 +352,7 @@ def get_tmp_filename(self, url):
 
     def fetch(self, url):
         net_arg = url
+        no_checksum = url[2] if isinstance(url, tuple) else False
```
shivan1b: Its usually nice to avoid these kinds of checks i.e. maintain the type of data all across the code so there is consistency.

vagishagupta23: @shivan1b I'm not sure if I understand this. Could you please elaborate more?

shivan1b: @vagishagupta23 checking multiple times in a code the "type" of a variable is usually not a good practice. You declare something as string then it should be string wherever its referenced, it should not be a tuple at some places and a string at others. Brings in data inconsistency and chances of code crashing increase.

vagishagupta23: @shivan1b Does that mean I should probably use some other check in place of this? But that seems difficult as url can possibly be tuple sometimes and string when added though the command line.
Or maybe instead of checking the type twice I could check it once, something like:
```
if isinstance(url, tuple):
    no_checksum = url[2]
    url = url[0]
else:
    no_checksum = False
    url = url
```
What are your thoughts on it?

shivan1b: @vagishagupta23 I meant that the internal functions should receive a consistent data type. As in if url is supposed to be a tuple, it should be passed as a tuple from every place that calls the inner function. Then we know that `fetch` always expects urls to be tuples and put checks accordingly inside the function.
Does this make things clear?

vagishagupta23: @shivan1b Ok, understood.

vagishagupta23:
A new PR is here #192

PR #193, opened 2019-09-05
Add no-checksum option

vagishagupta23:
Added a `--no-checksum` option to the add-source command and an optional
"no_checksum" in SourceConfiguration class for add-sources and
enabled-sources.
Also, a check is added to skip downloading the checksum URL if the
source is configured with no-checksum true.

Redmine issue:
    https://redmine.openinfosecfoundation.org/issues/3100

-- thank you.

  contribution agreement at
  changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3100

### DIFF
--- a/doc/add-source.rst
+++ b/doc/add-source.rst
@@ -25,6 +25,10 @@ Options
 
      add-source --http-header "X-API-Key: 1234"
 
+.. option:: --no-checksum
+
+   Skips downloading the checksum URL for the rule source.
+
 Common Options
 ==============
 
--- a/suricata/update/commands/addsource.py
+++ b/suricata/update/commands/addsource.py
@@ -28,13 +28,17 @@ except:
 
 logger = logging.getLogger()
 
+
 def register(parser):
     parser.add_argument("name", metavar="<name>", nargs="?",
                         help="Name of source")
     parser.add_argument("url", metavar="<url>", nargs="?", help="Source URL")
     parser.add_argument("--http-header", metavar="<http-header>",
                         help="Additional HTTP header to add to requests")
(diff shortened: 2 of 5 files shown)


jasonish:
Great, seems to be fixed up now.  However, we've decided to make a last minute change to the index format.  Instead of `no-checksum: true`, we'll use `checksum: false`. We will want to make the change for sources added via `add-source` as well, then invert all the logic.

Do you have time or want to do this? If not, I can use the work you've done here and inverse the logic.

vagishagupta23:
> Great, seems to be fixed up now. However, we've decided to make a last minute change to the index format. Instead of `no-checksum: true`, we'll use `checksum: false`. We will want to make the change for sources added via `add-source` as well, then invert all the logic.
> 
> Do you have time or want to do this? If not, I can use the work you've done here and inverse the logic.

No problem. I would take up this task and complete it. Asking just to be sure, we need to add a `--checksum` option to the add-source having `action="store_fault"` right ?

jasonish:
> No problem. I would take up this task and complete it. Asking just to be sure, we need to add a `--checksum` option to the add-source having `action="store_false"` right ?

I would keep the command line argument as `--no-checksum`, it makes sense on the command line.  But that would result in adding:
```
checksum: false
```
to the configuration file.

If `--no-checksum` is not provided, then there should be no `checksum` key in the configuration file. Which I think is the case now.

vagishagupta23:
Created a new PR with the suggested changes here #194.

PR #195, opened 2019-09-11
Add "check-versions" command

vagishagupta23:
Add a `suricata-update check-versions` subcommand that checks
the version of suricata and logs if the versions are up to
date, outdated or EOL.

Redmine issue:
    https://redmine.openinfosecfoundation.org/issues/2341

-- thank you.

  contribution agreement at
  changes made (if applicable)


suricata/update/main.py
```diff
@@ -1281,7 +1284,9 @@ def _main():
     suricata.update.net.set_user_agent_suricata_version(suricata_version.full)
 
     if args.subcommand:
-        if hasattr(args, "func"):
+        if args.subcommand == "check-versions" and hasattr(args, "func"):
+            return args.func(suricata_version)
```
shivan1b: Maybe we should not alter the existing subcommand structure. Let it be having functions that do not take any arguments. The code is written in a manner that we can always get args or configuration values on demand.
Probably it would be nice to bring the entire block of code where `suricata_version` is evaluated into some other file.
@jasonish is it OK to move the version evaluation code in `version.py`?

suricata/update/sources.py
```diff
@@ -130,6 +130,12 @@ def get_source_by_name(self, name):
             return self.index["sources"][name]
         return None
 
+    def get_versions(self):
+        try:
+            return self.index["versions"]
+        except KeyError:
+            raise Exception("Version information not in index")
```
shivan1b: Gently catch this exception and log an error? This gives an ugly traceback, maybe we should do it better.

suricata/update/commands/checkversions.py
```diff
@@ -0,0 +1,58 @@
...
+    parser.set_defaults(func=check_version)
+
+
+def check_version(suricata_version):
+    index_filename = sources.get_index_filename()
+    if not os.path.exists(index_filename):
+        logger.warning("No index exists, will use bundled index.")
+        logger.warning("Please run suricata-update update-sources.")
+    index = sources.Index(index_filename)
+    version = index.get_versions()
```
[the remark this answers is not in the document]
shivan1b: Chances of happening are low since we shall be providing the index but since the  called function is allowed to fail, we should probably handle all cases.

suricata/update/commands/checkversions.py
```diff
@@ -0,0 +1,58 @@
...
+    recommended = engine.parse_version(version['suricata']['recommended'])
```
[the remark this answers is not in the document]
shivan1b: Chances of happening are low since we shall be providing the index but since the  called function is allowed to fail, we should probably handle all cases.

shivan1b:
Also, this should probably be documented like the other subcommands: https://suricata-update.readthedocs.io/en/latest/update-sources.html

shivan1b:
@vagishagupta23 thanks a lot for your work. I'll be picking it up from here to do all the changes I have requested above since we are trying to include this in our next release therefore quite strict on schedule.

suricata/update/main.py
shivan1b: Leaving this as-is since extracting the version info from args seems to add a cyclic dependency in case the code is moved out.

shivan1b:
Thanks, Vagisha! Your commits from this PR are in #199

PR #197, opened 2019-10-02
Disable ja3_hash rules as needed - v1

jasonish:
Disable ja3_hash rules if Suricata not configured for ja3.

Rules using ja3_hash with fail to load if Suricata is not built with NSS,
or ja3 fingerprints are disabled.

Take into account the Suricata version as well, as not defining the
ja3_fingerprint configuration field in 5.0+ will leave it enabled, but in
older versions, it will remain disabled if not defined.

https://redmine.openinfosecfoundation.org/issues/3215

### DIFF
--- a/suricata/update/config.py
+++ b/suricata/update/config.py
@@ -208,9 +208,15 @@ def init(args):
                 0, os.path.join(
                     build_info["sysconfdir"], "suricata/suricata.yaml"))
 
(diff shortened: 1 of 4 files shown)


jasonish:
Closing. Updated PR here:
https://github.com/OISF/suricata-update/pull/198

PR #210, opened 2019-11-12
Support for rule filtering based on metadata key-value pairs

whartond:
  contribution agreement at
- [X] I have updated the user guide (in doc/userguide/) to reflect the
  changes made (if applicable)

Rule Matching: Rules can be enabled and disabled based on metadata key-value pairs
present in the Suricata rule metadata keyword.

    - Added Rule Matching support for filtering rulesets based on metadata key-value pairs,
      using Boolean filters.

    - New conf file, 'metadata.conf' (functionally empty by default); passing conf file on
      command line also supported.

    - Leverages BETTER Schema standard (https://better-schema.readthedocs.io/).

    - Added requirement, 'aristotle'; code additions leverage use the Aristotle library
      (https://github.com/secureworks/aristotle/, https://pypi.org/project/aristotle/).

    - Updated docs to include new feature(s).

### DIFF
--- a/doc/update.rst
+++ b/doc/update.rst
@@ -186,6 +186,7 @@ modification can be done with the following:
 - regular expression
 - rule group
 - filename
+- metadata filter
 
 Signature ID Matching
 ---------------------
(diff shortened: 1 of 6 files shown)


jasonish:
Belated thanks for the contribution. Its been on my mind for a while.

While I generally like the idea of a more standardized schema, we can't take this change in at this time. This is mostly due to this PR requiring modules that are not generally available across all package management systems (being in pypi isn't enough), and we try to keep Suricata-Update free of any external dependencies with the exception of the Yaml module for obvious reasons.  A PR such as this should somehow be optional as it uses some extra libraries and introduces a new syntax for matching rules.  I imagine having some sort of post-processing plugin support at some point that would allow for something like this, where people can opt-in. But I'm not sure how that will fit into our planning at this time.

jasonish:
Closing due to library requirements.

PR #211, opened 2019-11-13
Disable SMB rules if not supported by Suricata - v1

jasonish:
If Suricata is less than 5, and Rust is not enabled, disable
smb rules as they require Rust.

Ticket 3280:
https://redmine.openinfosecfoundation.org/issues/3280


jasonish:
Merged.

PR #218, opened 2020-01-06
move modify into loop with others and skip format check

ct0br0:
Moved the modify part into the loop above to - seems to run way faster (15min instead of an hour and a half).

May be missing something, but removed the "if new_rule.format() != rule.format()" because it seemed like extra work to me. If a user has a modify that's messed up and produces the same rule all this does is run an if (in some cases 5000+ times) more, on top of the already poor rule running for nothing.

didn't do this - lmk if you want me to

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -1063,12 +1063,10 @@ def _main():
                 rulemap[rule.id] = fltr.run(rule)
                 drop_count += 1
 
-    # Apply modify filters.
-    for fltr in modify_filters:
-        for key, rule in rulemap.items():
+        for fltr in modify_filters:
             if fltr.match(rule):
                 new_rule = fltr.run(rule)
-                if new_rule and new_rule.format() != rule.format():
+                if new_rule:
                     rulemap[rule.id] = new_rule
                     modify_count += 1
 


jasonish:
> Moved the modify part into the loop above to - seems to run way faster (15min instead of an hour and a half).

Can you provide an example configuration that takes so long to run... So we can verify?

ct0br0:
any large count of "sid what replace" should work
we have roughly 5000 "sid classification:this classification:that"

instead of redoing the rules (60k) per 5k modifies - do 5k checks on a modify while checking all the rules

ct0br0:
were you ever able to test this?

jasonish:
In my contrived tests I'm not seeing much of a difference, but I probably should have written it like this in the first place.

Yes, do the contributor agreement. Thanks!

ct0br0:
awesome. just did it. hope youre doing well

suricata/update/main.py
```diff
@@ -1063,12 +1063,10 @@ def _main():
                 rulemap[rule.id] = fltr.run(rule)
                 drop_count += 1
 
-    # Apply modify filters.
-    for fltr in modify_filters:
-        for key, rule in rulemap.items():
+        for fltr in modify_filters:
             if fltr.match(rule):
                 new_rule = fltr.run(rule)
-                if new_rule and new_rule.format() != rule.format():
```
inashivb, line 1071: I may be missing something here but it seems like the modify filter can return two values:
1. Original rule in case the filter resulted in an invalid rule after modification.
2. Modified rule in case everything happened successfully.

This check made sure that the modified rule was valid and only a valid rule gets added to the rulemap.
@jasonish is this OK?

jasonish: This was a way to detect failed modify filters adding to the count, and I think its quite an expensive check to be doing now that I'm seeing more examples of people modifying 1000s, or even 10000+ rules. A log error is still printed. So I think its OK.

jasonish:
Merged. Thanks!

PR #219, opened 2020-03-23

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2688

Note:
- Does not handle unmerged rule files
- Does not handle duplicate filenames

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -246,7 +246,6 @@ def load_filters(filename):
     return filters
 
 def load_drop_filters(filename):
-    
     matchers = load_matchers(filename)
     filters = []
 
@@ -374,6 +373,7 @@ def write_merged(filename, rulemap):
                     removed.append(rule)
                 elif rule.format() != rulemap[rule.id].format():
                     modified.append(rulemap[rule.id])
+
         for key in rulemap:
             if not key in oldset:
                 added.append(key)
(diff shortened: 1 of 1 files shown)


suricata/update/main.py
```diff
@@ -387,10 +387,22 @@ def write_merged(filename, rulemap):
...
+                load_attr = [el for el in rule.dataset.split(",") if "load" in el][0]
+                dataset_fname = load_attr.split(" ")[1] # ASK if there can only be one filename
+                logger.debug("Copying dataset file %s to output directory" % dataset_fname)
+                shutil.copy(os.path.join(config.get_cache_dir(), dataset_fname),
+                            os.path.join(config.get_output_dir(), dataset_fname))
+            if "filemd5" in rule:
+                filemd5_fname = rule.get("filemd5")
+                logger.debug("Copying filemd5 file %s to output directory" % filemd5_fname)
+                shutil.copy(os.path.join(config.get_cache_dir(), filemd5_fname),
+                            os.path.join(config.get_output_dir(), filemd5_fname))
```
inashivb: @jasonish here I am assuming that the file exists in the cache dir alongwith its rule file but the [issue](https://redmine.openinfosecfoundation.org/issues/2688) says that the file is not moved from the [default](https://suricata.readthedocs.io/en/latest/rules/file-keywords.html#filemd5) mentioned location i.e. `/etc/suricata/rules` to the rule directory.
Is it wrong to use cache directory here?

suricata/update/main.py
```diff
@@ -387,10 +387,22 @@ def write_merged(filename, rulemap):
...
                         len(modified)))
-    
     with io.open(filename, encoding="utf-8", mode="w") as fileobj:
-        for rule in rulemap:
-            print(rulemap[rule].format(), file=fileobj)
+        for sid in rulemap:
+            rule = rulemap[sid]
+            if "dataset" in rule:
+                load_attr = [el for el in rule.dataset.split(",") if "load" in el][0]
+                dataset_fname = load_attr.split(" ")[1] # ASK if there can only be one filename
```
inashivb: Could there be just one filename with load param?

suricata/update/main.py
jasonish: For this you can ignore the idea of a default location.  The default location will be where Suricata-Update decides to put these files.

Right now if you use such a ruleset, these data files are just ignored, and never extracted. So the fix is to extract them and put them alongside the rules (then handle all the edge cases that come up along with that).

suricata/update/main.py
jasonish: Yes, I think its OK to assume this.

jasonish:
I also vote we kill the travis-ci tests, I don't want to debug this one, and the GitHub CI ones were made to pickup some issues not being found by Travis anyways.

suricata/update/main.py
jasonish: So the file does not exist in the cache directory, it is inside of the archive file along side the rule file.  Say my download rule file is rules.tar.gz, inside I might see something like:
- rules/filemd5.rules (the rule file)
- rules/filemd5.list (the md5 list)

I'll attach a sample rule archive that I've created for testing to this PR.

jasonish:
[rules.tar.gz](https://github.com/OISF/suricata-update/files/4382295/rules.tar.gz)

suricata/update/main.py
inashivb: Thank you very much!

inashivb:
Replaced with #220

PR #212, opened 2019-11-13
Ticket 3174 - Handle duplicate filenames - v1

jasonish:
Handle duplicate filenames across sources.
    
    In the dict that holds the downloaded files, prefix the filename
    with the URL using '!' as a separator.
    
    Ticket 3174:
    https://redmine.openinfosecfoundation.org/issues/3174

PR #220, opened 2020-03-26

inashivb:
https://redmine.openinfosecfoundation.org/issues/2688

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -246,7 +246,6 @@ def load_filters(filename):
     return filters
 
 def load_drop_filters(filename):
-    
     matchers = load_matchers(filename)
     filters = []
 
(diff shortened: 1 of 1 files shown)


inashivb:
@jasonish seems like rebasing this PR over #212  would take care of duplicate names for the files as well? Please let me know if this is an incorrect interpretation.

jasonish:
Ok, the basics now work, but needs to expanded:

- The filemd5 handling of data files should be made generic over filemd5, filesha1 and filesha256.
- Handle datafiles in a subdirectory, for example if the rule has `filemd5:datafiles/md5.list`, and the rule archive contains `rules/datafiles/md5.list` it should be found

Issues to think about:
- If 2 sources include `rules/datafiles/md5.list` one will be lost when the rules are loaded into the dict. PR #212 will solve that problem. But this feature will need to handle the conflict when writing out the files to disk.
- Removing files from the rule directory when they are no longer referenced by any rules.

jasonish:
My latest test rule file:
[rules.tar.gz](https://github.com/OISF/suricata-update/files/4409833/rules.tar.gz)

inashivb:
Thanks, Jason. Please check the new PR #222

PR #221, opened 2020-03-31
Ticket 3174 - Duplicate filename handling - v2

jasonish:
Previous PR: https://github.com/OISF/suricata-update/pull/212

Changes from last PR:
- Remove the URL scheme from the internal filename format to avoid
  links from being clickable.
- Remove Python 3.4 tests due to deprecations.

https://redmine.openinfosecfoundation.org/issues/3174

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -205,11 +205,22 @@ class Fetch:
         if url:
             try:
                 fetched = self.fetch(url)
-                files.update(fetched)
+                for key in fetched:
+                    safe_filename = self.safe_filename(url[0], key)
+                    files[safe_filename] = fetched[key]
             except URLError as err:
                 url = url[0] if isinstance(url, tuple) else url
                 logger.error("Failed to fetch %s: %s", url, err)
         return files
+
+    def safe_filename(self, url, filename):
+        """Create a safe/conflict free filename from the URL and the full path
+        of a file inside the archive.
+        """
+        scheme_end = url.find("://")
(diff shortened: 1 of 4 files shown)


suricata/update/main.py
```diff
@@ -199,21 +199,29 @@ def fetch(self, url):
...
 
+    def safe_filename(self, url, filename):
+        """Create a safe/conflict free filename from the URL and the full path
+        of a file inside the archive.
+        """
+        scheme_end = url.find("://")
+        if scheme_end > 0:
+            return "{}!{}".format(url[scheme_end + 1:], filename)
+        return "{}!{}".format(url, filename)
+
```
inashivb, line 224: This works well. I have a few questions though. I see debug info like:
```
1/4/2020 -- 17:03:00 - <Debug> -- Parsing //rules.emergingthreats.net/open/suricata-5.0.1/emerging.rules.tar.gz!rules/emerging-web_client.rules.
```
which is not clickable any more but do we want the front `//` in the file name?

Also, I see that in the end with `--no-merge` option, some files written in `/var/lib/suricata/rules` are named `trafficid.rules!trafficid.rules`, do we want this nomenclature?

jasonish: To me I see '//' as this came from something over a URL.. But could remove it.

Have the leading text including the '!' should never be in the output. I'll have to fix that. Thanks for catching it.

jasonish:
Replaced by https://github.com/OISF/suricata-update/pull/240.

PR #222, opened 2020-04-03

inashivb:
https://redmine.openinfosecfoundation.org/issues/2688

Notes from previous PR by Jason:

> Issues to think about:
>    - If 2 sources include rules/datafiles/md5.list one will be lost when the rules are loaded into the dict. PR #212 will solve that problem. But this feature will need to handle the conflict when writing out the files to disk.
>  -   Removing files from the rule directory when they are no longer referenced by any rules

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -92,6 +92,9 @@ DEFAULT_OUTPUT_RULE_FILENAME = "suricata.rules"
 
 INDEX_EXPIRATION_TIME = 60 * 60 * 24 * 14
 
+# Rule keywords that come with files
+file_kw = ["filemd5", "filesha1", "filesha256", "dataset"]
+
 class Fetch:
 
     def __init__(self):
@@ -246,7 +249,6 @@ def load_filters(filename):
     return filters
 
 def load_drop_filters(filename):
-    
     matchers = load_matchers(filename)
     filters = []
 
@@ -356,7 +358,28 @@ def load_dist_rules(files):
                 logger.error("Failed to open %s: %s" % (path, err))
                 sys.exit(1)
 
-def write_merged(filename, rulemap):
+def handle_dataset_files(rule, dep_files):
+    load_attr = [el for el in rule.dataset.split(",") if "load" in el][0]
+    dataset_fname = os.path.basename(load_attr.split(" ")[1])
+    filename = [fname for fname, content in dep_files.items() if fname == dataset_fname]
+    if filename:
+        logger.debug("Copying dataset file %s to output directory" % dataset_fname)
+        with open(os.path.join(config.get_output_dir(), dataset_fname), "w+") as fp:
+            fp.write(dep_files[dataset_fname].decode("utf-8"))
+    else:
+        logger.error("Dataset file %s was not found" % dataset_fname)
+
+def handle_filehash_files(rule, dep_files, fhash):
+    filehash_fname = os.path.basename(rule.get(fhash))
+    filename = [fname for fname, content in dep_files.items() if fname == filehash_fname]
+    if filename:
+        logger.debug("Copying %s file %s to output directory" % (fhash, filehash_fname))
+        with open(os.path.join(config.get_output_dir(), filehash_fname), "w+") as fp:
+            fp.write(dep_files[filehash_fname].decode("utf-8"))
+    else:
+        logger.error("%s file %s was not found" % (fhash, filehash_fname))
+
+def write_merged(filename, rulemap, dep_files):
 
     if not args.quiet:
         # List of rule IDs that have been added.
@@ -374,6 +397,7 @@ def write_merged(filename, rulemap):
                     removed.append(rule)
                 elif rule.format() != rulemap[rule.id].format():
                     modified.append(rulemap[rule.id])
+
         for key in rulemap:
             if not key in oldset:
                 added.append(key)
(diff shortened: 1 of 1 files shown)


jasonish:
Given my rule archive that looks like (attached)
```
rules/filemd5.list
rules/datafiles
rules/datafiles/filesha256.list
rules/datafiles/filesha1.list
rules/test.rules
```
the output of Suricata-Update looks like:
```
data/rules/filesha256.list
data/rules/filesha1.list
data/rules/suricata.rules
data/rules/filemd5.list
```
The `list` files are in the same directory as the rule file, but the rules are still looking for the files in a `datafiles` directory. Suricata will fail to load the rules with an error like:
```
[1407815] 8/4/2020 -- 11:16:20 - (detect-file-hash-common.c:249) <Error> (DetectFileHashParse) -- [ERRCODE: SC_ERR_OPENING_RULE_FILE(41)] - opening hash file ../../suricata-update/datasets/data/datafiles/filesha1.list: No such file or directory
```

So the output directory needs to respect the input directory layout for these datafiles.

inashivb:
> Given my rule archive that looks like (attached)
> 
> ```
> rules/filemd5.list
> rules/datafiles
> rules/datafiles/filesha256.list
> rules/datafiles/filesha1.list
> rules/test.rules
> ```
> 
> the output of Suricata-Update looks like:
> 
> ```
> data/rules/filesha256.list
> data/rules/filesha1.list
> data/rules/suricata.rules
> data/rules/filemd5.list
> ```
> 
> The `list` files are in the same directory as the rule file, but the rules are still looking for the files in a `datafiles` directory. Suricata will fail to load the rules with an error like:
> 
> ```
> [1407815] 8/4/2020 -- 11:16:20 - (detect-file-hash-common.c:249) <Error> (DetectFileHashParse) -- [ERRCODE: SC_ERR_OPENING_RULE_FILE(41)] - opening hash file ../../suricata-update/datasets/data/datafiles/filesha1.list: No such file or directory
> ```
> 
> So the output directory needs to respect the input directory layout for these datafiles.

I see. Please let me know if I understood correctly, you are suggesting to update the rules with the absolute path that we are moving the file to. Is that correct?

jasonish:
> I see. Please let me know if I understood correctly, you are suggesting to update the rules with the absolute path that we are moving the file to. Is that correct?

No, I'd just make sure the list files get put into a directory where the rule expects to find them. Shouldn't need to update that rule itself.

We need to avoid absolute paths as someone may run Suricata-Update, package up the output directory and push it out to their sensors where the paths may be different.

So the output directory in this case should look like:
```
data/rules/datafiles/filesha256.list
data/rules/filesha1.list
data/rules/suricata.rules
data/rules/datafiles/filemd5.list
```

We may have to get into rule rewriting the rules to avoid filename clashes.. For example, if 2 rules from different sources include a data file of the same name, but are different. Then we'll have to rename the datafiles and update the rules. But will still keep a relative path.

inashivb:
> > I see. Please let me know if I understood correctly, you are suggesting to update the rules with the absolute path that we are moving the file to. Is that correct?
> 
> No, I'd just make sure the list files get put into a directory where the rule expects to find them. Shouldn't need to update that rule itself.
> 
> We need to avoid absolute paths as someone may run Suricata-Update, package up the output directory and push it out to their sensors where the paths may be different.
> 
> So the output directory in this case should look like:
> 
> ```
> data/rules/datafiles/filesha256.list
> data/rules/filesha1.list
> data/rules/suricata.rules
> data/rules/datafiles/filemd5.list
> ```
> 
> We may have to get into rule rewriting the rules to avoid filename clashes.. For example, if 2 rules from different sources include a data file of the same name, but are different. Then we'll have to rename the datafiles and update the rules. But will still keep a relative path.

Thanks, Jason!
I see on Suricata test that the dir expected is
```
3/5/2020 -- 02:06:42 - <Error> -- [ERRCODE: SC_ERR_OPENING_RULE_FILE(41)] - opening hash file /var/lib/suricata/datafiles/filesha1.list: No such file or directory
```
It seems to be skipping `/rules/`. state_dir + whatever was in the rule seems to be the path where Suricata looks into. Is this the right path or am I doing something wrong?

jasonish:
> It seems to be skipping `/rules/`. state_dir + whatever was in the rule seems to be the path where Suricata looks into. Is this the right path or am I doing something wrong?

Looks like an issue with Suricata. When using -S, and the rule file falls outside of default-rule-path, it can't find the datafiles.  We can probably fix this with some --set command line options.

But first, I'm stuck with these errors when I run suricata-update against my test rule archive:
```
28/5/2020 -- 11:38:01 - <Error> -- filemd5 file fileextraction-chksum.list was not found
28/5/2020 -- 11:38:01 - <Error> -- filesha1 file fileextraction-chksum.list was not found
28/5/2020 -- 11:38:01 - <Error> -- filesha256 file fileextraction-chksum.list was not found
```
My command line is:
```
./bin/suricata-update -D ./data --url file:///home/jason/oisf/code/suricata-update/rules.tar.gz
```
and rule
[rules.tar.gz](https://github.com/OISF/suricata-update/files/4697326/rules.tar.gz)
s.tar.gz is attached.

inashivb:
> > It seems to be skipping `/rules/`. state_dir + whatever was in the rule seems to be the path where Suricata looks into. Is this the right path or am I doing something wrong?
> 
> Looks like an issue with Suricata. When using -S, and the rule file falls outside of default-rule-path, it can't find the datafiles. We can probably fix this with some --set command line options.
> 
> But first, I'm stuck with these errors when I run suricata-update against my test rule archive:
> 
> ```
> 28/5/2020 -- 11:38:01 - <Error> -- filemd5 file fileextraction-chksum.list was not found
> 28/5/2020 -- 11:38:01 - <Error> -- filesha1 file fileextraction-chksum.list was not found
> 28/5/2020 -- 11:38:01 - <Error> -- filesha256 file fileextraction-chksum.list was not found
> ```
> 
> My command line is:
> 
> ```
> ./bin/suricata-update -D ./data --url file:///home/jason/oisf/code/suricata-update/rules.tar.gz
> ```
> 
> and rule
> [rules.tar.gz](https://github.com/OISF/suricata-update/files/4697326/rules.tar.gz)
> s.tar.gz is attached.

Thanks, Jason. Please see the updated PR #234 
Please note that it only covers the full functionality for filemd5. datasets have not been tested yet. The rule archive you provided seems to be passing with that but it still has the issue of `/rules/` missing in the paths.

PR #225, opened 2020-04-10
Better environment detection for user-agent - v1

jasonish:
Don't rely on Python's platform.linux_distribution module as it
has been deprecated and removed in the latest version of Python.

Instead use more generic ways to pick up the distribution such
as looking at /etc/os-release, or falling back to uname.

Should also pickup the version of FreeBSD, which was previously
missing.


jasonish:
Merged.

PR #236, opened 2020-06-19
manage classification.config

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3203

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -357,6 +357,56 @@ def load_dist_rules(files):
                 logger.error("Failed to open %s: %s" % (path, err))
                 sys.exit(1)
 
+def load_classification(suriconf, files):
+    filename = os.path.join("suricata", "classification.config")
+    dirs = []
+    classification_arr = {}
+    if "sysconfdir" in suriconf.build_info:
+        dirs.append(os.path.join(suriconf.build_info["sysconfdir"], filename))
+    if "datarootdir" in suriconf.build_info:
+        dirs.append(os.path.join(suriconf.build_info["datarootdir"], filename))
+
+    for path in dirs:
+        if os.path.exists(path):
+            with open(path) as fp:
+                for line in fp:
+                    if line.startswith("#") or not line.strip():
+                        continue
+                    config_classification = line.split(":")[1].strip()
+                    key, desc, priority = config_classification.split(",")
+                    if key in classification_arr:
+                        if classification_arr[key][1] >= priority:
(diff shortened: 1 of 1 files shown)


suricata/update/main.py
```diff
@@ -357,6 +357,56 @@ def load_dist_rules(files):
...
+            with open(path) as fp:
+                for line in fp:
+                    if line.startswith("#") or not line.strip():
+                        continue
+                    config_classification = line.split(":")[1].strip()
+                    key, desc, priority = config_classification.split(",")
+                    if key in classification_arr:
+                        if classification_arr[key][1] >= priority:
+                            continue
+                    classification_arr[key] = [desc, priority, line.strip()]
```
jasonish, line 380: Some debug logging line is needed here.  When I run with `-v` I should be able to see what classification.config its loading off my file system.

inashivb:
> We have #221 to help address this. But I was thinking store the loaded files in a list instead of a dict might be a more straightforward way to handle this.
> 
Do you mean a list of classification files? Would this work? https://github.com/OISF/suricata-update/pull/236/files#diff-b117dc9956be2a723a248660048c4a4bR1080

jasonish:
> Do you mean a list of classification files? Would this work? https://github.com/OISF/suricata-update/pull/236/files#diff-b117dc9956be2a723a248660048c4a4bR1080

Was thinking a list for all files. Not sure if that would work, cause its working on the "dict" of files loaded right? In the case, duplicate files by name (not content) are already lost.

jasonish:
> Was thinking a list for all files. Not sure if that would work, cause its working on the "dict" of files loaded right? In the case, duplicate files by name (not content) are already lost.

Outside of the scope of this PR though.

inashivb:
> > Do you mean a list of classification files? Would this work? https://github.com/OISF/suricata-update/pull/236/files#diff-b117dc9956be2a723a248660048c4a4bR1080
> 
> Was thinking a list for all files. Not sure if that would work, cause its working on the "dict" of files loaded right? In the case, duplicate files by name (not content) are already lost.

We could use hash of the files as keys instead of the name. Would that make sense?

jasonish:
> We could use hash of the files as keys instead of the name. Would that make sense?

Yeah, it might.  I have to jump back into #221 and solve this issue for good.

inashivb:
Replaced with #239

PR #238, opened 2020-06-26
Small fixups/v1

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) tickets:
https://redmine.openinfosecfoundation.org/issues/3262
https://redmine.openinfosecfoundation.org/issues/3550
https://redmine.openinfosecfoundation.org/issues/3535

### DIFF
--- a/suricata/update/engine.py
+++ b/suricata/update/engine.py
@@ -26,7 +26,9 @@ import os.path
 import subprocess
 import re
 import logging
+import shutil
 import yaml
+import tempfile
 from collections import namedtuple
 
 logger = logging.getLogger()
@@ -163,10 +165,11 @@ def get_version(path):
 
 def test_configuration(suricata_path, suricata_conf=None, rule_filename=None):
     """Test the Suricata configuration with -T."""
+    tempdir = tempfile.TemporaryDirectory()
     test_command = [
         suricata_path,
         "-T",
-        "-l", "/tmp",
+        "-l", tempdir.name,
     ]
     if suricata_conf:
         test_command += ["-c", suricata_conf]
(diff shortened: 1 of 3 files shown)


suricata/update/engine.py
```diff
@@ -163,10 +165,11 @@ def get_version(path):
 
 def test_configuration(suricata_path, suricata_conf=None, rule_filename=None):
     """Test the Suricata configuration with -T."""
+    tempdir = tempfile.TemporaryDirectory()
```
jasonish, line 168: This method does not appear to be available on Python 2. `tempfile.mkdtemp` is an option that should work on both.

suricata/update/engine.py
```diff
@@ -178,9 +181,12 @@ def test_configuration(suricata_path, suricata_conf=None, rule_filename=None):
 
     logger.debug("Running %s; env=%s", " ".join(test_command), str(env))
     rc = subprocess.Popen(test_command, env=env).wait()
-    if rc == 0:
-        return True
-    return False
+    ret = True if rc == 0 else False
+
+    # Cleanup the temp dir
+    shutil.rmtree(tempdir.name)
```
jasonish, line 187: This causes a crash for me as the tempdir tries to remove itself on destruction, but its already gone.  I think `tempdir.cleanup()` will fix this instead of using rmtree.  But again, not available on Python 2 which I'd still like to support where possible.

inashivb: I do not see a `cleanup` function in the instance of tempdir (with `mkdtemp`). Should I just wrap this rmtree call in a try except block to avoid getting a traceback in such cases?

jasonish: > I do not see a `cleanup` function in the instance of tempdir (with `mkdtemp`). Should I just wrap this rmtree call in a try except block to avoid getting a traceback in such cases?

The tempdir.cleanup() only exists on Python 3 with tempfile.TemporaryDirectory().  The crash comes from within the module, not your code.

Best option is to use mkdtemp, and then keep the manual cleanup that you now, as mkdtemp doesn't have cleanup built-in.

inashivb:
Replaced with #247

PR #240, opened 2020-07-08
Duplicate filename handling - v3

jasonish:
Previous PR: https://github.com/OISF/suricata-update/pull/221

Changes from last PR:
- Load files into a list instead of trying to come up with a filename scheme to avoid duplication.

By loading all downloaded rule files into the same dict, filenames
that are the same cause other files to be lost, and the content
may be different.

Instead use of list objects that tracks the filename and the content
to avoid losing the contents of a filename that already exists.

If the contents is duplicated, the rule deduplication process
will catch that.

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/3174

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -31,6 +31,7 @@ import glob
 import io
 import tempfile
 import signal
+from collections import namedtuple
 
 try:
     # Python 3.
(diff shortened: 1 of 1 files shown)


jasonish:
I think this is a little more important now with the classification.config support merged, as that file is likely to have the same name accross multiple rulesets.

inashivb (approved):
LGTM

inashivb:
Ready for review as a part of #247

PR #244, opened 2020-08-22
dataset parsing: handle spaces between "," - v1

jasonish:
Fix dataset parsing so the filename can be parsed from:
    
  dataset:isset, sslbl, type md5, load sslbl-fingerprints.md5;


inashivb (approved):
LGTM

inashivb:
Ready for review as a part of #247

PR #257, opened 2020-10-08
Add warning to update sources if no index found

TharushiJay:
-- thank you.

  contribution agreement at
  changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3249

Describe changes:
-  Similar to enable-source and check-versions commands that ask the user to run update-sources, added the fix in list-sources command to ask the user to run the update-sources command instead of updating the sources without asking the user.

### DIFF
--- a/suricata/update/commands/listsources.py
+++ b/suricata/update/commands/listsources.py
@@ -73,7 +73,6 @@ def list_sources():
     if not sources.source_index_exists(config):
         logger.warning("Source index does not exist, will use bundled one.")
         logger.warning("Please run suricata-update update-sources.")
-        return 1
 
     index = sources.load_source_index(config)
     for name, source in index.get_sources().items():
--- a/suricata/update/commands/listsources.py
+++ b/suricata/update/commands/listsources.py
@@ -73,6 +73,7 @@ def list_sources():
     if not sources.source_index_exists(config):
         logger.warning("Source index does not exist, will use bundled one.")
         logger.warning("Please run suricata-update update-sources.")
+        return 1
 
     index = sources.load_source_index(config)
     for name, source in index.get_sources().items():
--- a/suricata/update/commands/listsources.py
+++ b/suricata/update/commands/listsources.py
@@ -21,7 +21,6 @@ import logging
 from suricata.update import config
 from suricata.update import sources
 from suricata.update import util
-from suricata.update.commands.updatesources import update_sources
 from suricata.update import exceptions
 
 logger = logging.getLogger()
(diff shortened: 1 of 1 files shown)


suricata/update/commands/listsources.py
```diff
@@ -72,11 +71,10 @@ def list_sources():
...
     free_only = config.args().free
     if not sources.source_index_exists(config):
-        logger.info("No source index found, running update-sources")
-        try:
-            update_sources()
-        except exceptions.ApplicationError as err:
-            logger.warning("%s: will use bundled index.", err)
+        logger.warning("Source index does not exist, will use bundled one.")
+        logger.warning("Please run suricata-update update-sources.")
+        return 1
```
jasonish, line 76: We don't need to return here. The rest of the code will run fine with the "bundled" index. It may be out of date, but we thats what this warning is for.

jasonish (changes requested):
Just a minor change required, otherwise looks good.

suricata/update/commands/listsources.py
TharushiJay: Thanks! I removed the return statement, but the checks are still failing.

jasonish (changes requested):
In order to keep our git history clean we prefer fixup commits to be squashed into the previous commit.  In this case, the last commit could have been removed. Another option would be to squash these 3 commits into one, and then create a new PR.  Sometimes with smaller changes like this starting with a fresh branch might be easier than dealing with git rebase.

With respect to this PR, all we really want to see is the first commit, with no sign of the return being added then removed.

Thanks!

suricata/update/commands/listsources.py
jasonish: > Thanks! I removed the return statement, but the checks are still failing.

Just looking at this now.  I think we have an integration test that checks for the index.yaml being downloaded after a list-sources, this will have to be changed - will confirm.

jasonish: Ok, to tests/integration_tests.py checks for the index.yaml at line 66, to verify that it was downloaded as part of a list source.  With this change, we should verify that it was *not* downloaded.

TharushiJay: Thanks for the information. Sorry for the multiple pull requests. Is there any chance I could run the checks locally before opening a pull request?

TharushiJay:
Replaced with #257

PR #260, opened 2020-10-09
Remove options not common to all commands

TharushiJay:
  contribution agreement at
  changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3250

Describe changes:

Limited the work to researching which options can be moved out of global arguments.

### DIFF
--- a/suricata/update/parsers.py
+++ b/suricata/update/parsers.py
@@ -56,13 +56,7 @@ global_arg = [
       'help': "Override Suricata version"}),
     (("--user-agent",),
      {'metavar': '<user-agent>',
-      'help': "Set custom user-agent string"}),
-    (("--no-check-certificate",),
-     {'action': 'store_true', 'default': None,
-      'help': "Disable server SSL/TLS certificate verification"}),
-    (("-V", "--version"),
-     {'action': 'store_true', 'default': False,
-      'help': "Display version"})
+      'help': "Set custom user-agent string"})
 ]
 
 # Update arguments - command line options for suricata-update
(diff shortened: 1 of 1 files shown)


suricata/update/parsers.py
```diff
@@ -141,6 +135,9 @@
     (("--offline",),
      {'action': 'store_true',
       'help': "Run offline using most recent cached rules"}),
+    (("-V", "--version"),
+     {'action': 'store_true', 'default': False,
+      'help': "Display version"}),
```
jasonish, line 140: I think its OK to move `-V` here, but the check for this flag where it prints the version and exits will need to be changed as its not longer in the global arguments.  For example, this command fails now:
```
./bin/suricata-update
```

suricata/update/parsers.py
```diff
@@ -56,13 +56,7 @@
       'help': "Override Suricata version"}),
     (("--user-agent",),
      {'metavar': '<user-agent>',
-      'help': "Set custom user-agent string"}),
-    (("--no-check-certificate",),
-     {'action': 'store_true', 'default': None,
-      'help': "Disable server SSL/TLS certificate verification"}),
```
jasonish, line 62: I don't see this re-added anywhere.  It either needs to remain global, or added to any command that may go out on the network.

jasonish (changes requested):

TharushiJay:
Closing this PR since it was advised to restrict the work to researching and finding the redundant options for all the commands.
Findings are presented at : https://redmine.openinfosecfoundation.org/issues/3250

PR #261, opened 2020-10-15
Cleanup help output

TharushiJay:
Cleaning up the suricata help output and hiding the extra options
behind a "--show-advanced" option.

Hidden options are:

* --user-agent <user-agent>
* --no-check-certificate
* --yaml-fragment <filename>
* --url <url>
* --local <path>
* --sid-msg-map <filename>
* --sid-msg-map-2 <filename>
* --ignore <pattern>
* --no-ignore
* --threshold-in <filename>
* --threshold-out <filename>
* --dump-sample-configs
* --etopen
* --reload-command <command>
* --test-command <command>
* --no-merge

  contribution agreement at
  changes made (if applicable)

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/3974

Describe changes:
-  Cleaned up the suricata-update help output, by hiding the extra options behind a "--show-advanced" option
-  Updated the user guide to add "-s", "--show-advanced" option

### DIFF
--- a/suricata/update/parsers.py
+++ b/suricata/update/parsers.py
@@ -31,6 +31,8 @@ except:
 
 default_update_yaml = config.DEFAULT_UPDATE_YAML_PATH
 
+show_advanced = False
+
 # Global arguments - command line options for suricata-update
 global_arg = [
     (("-v", "--verbose"),
@@ -56,13 +58,18 @@ global_arg = [
       'help': "Override Suricata version"}),
     (("--user-agent",),
      {'metavar': '<user-agent>',
-      'help': "Set custom user-agent string"}),
+      'help': "Set custom user-agent string" 
+      if show_advanced else argparse.SUPPRESS}),
     (("--no-check-certificate",),
      {'action': 'store_true', 'default': None,
-      'help': "Disable server SSL/TLS certificate verification"}),
+      'help': "Disable server SSL/TLS certificate verification"
+      if show_advanced else argparse.SUPPRESS}),
     (("-V", "--version"),
      {'action': 'store_true', 'default': False,
-      'help': "Display version"})
+      'help': "Display version"}),
+    (("-s","--show-advanced"),
+     {'action': 'store_true',
+      'help': "Show advanced options"}),
 ]
 
 # Update arguments - command line options for suricata-update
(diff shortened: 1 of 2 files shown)


suricata/update/parsers.py
```diff
@@ -239,3 +233,34 @@ def parse_arg():
...
+def show_all_help():
+  update_arg[2][1]['help'] = "Output YAML fragment for rule inclusion"
+  update_arg[5][1]['help'] = "Generate a sid-msg.map file"
+  update_arg[6][1]['help'] = "Generate a v2 sid-msg.map file"
+  update_arg[11][1]['help'] = "Filenames to ignore (can be specified multiple times; default: *deleted.rules)"
+  update_arg[12][1]['help'] = "Disables the ignore option."
+  update_arg[13][1]['help'] = "Filename of rule thresholding configuration"
+  update_arg[14][1]['help'] = "Output of processed threshold configuration"
+  update_arg[15][1]['help'] = "Dump sample config files to current directory"
+  update_arg[21][1]['help'] = "Do not merge the rules into a single file"
```
jasonish, line 247: I think this will be fragile against changes to the arguments - the indexes might get out of sync. I had something more in mind like:
```
show_advanced = False
```
```
     (("--sid-msg-map",),
-     {'metavar': '<filename>', 'help': argparse.SUPPRESS}),
+     {'metavar': '<filename>', 'help': "Generate a sid-msg.map file" if show_advanced else argparse.SUPPRESS}),
     (("--sid-msg-map-2",),
```

I guess there is the issue of setting `show_advanced` to True. This will probably have to be done before parsing any args with a quick scan through argv..
```
if "-s" in sys.argv or "--show-all-options" in sys.argv:
    show_advanced = True
```

TharushiJay: That's a great way to solve the issue!

suricata/update/parsers.py
```diff
@@ -56,13 +61,17 @@
...
+      'help': "Set custom user-agent string" 
+      if show_advanced else argparse.SUPPRESS}),
     (("--no-check-certificate",),
      {'action': 'store_true', 'default': None,
       'help': "Disable server SSL/TLS certificate verification"}),
     (("-V", "--version"),
      {'action': 'store_true', 'default': False,
-      'help': "Display version"})
+      'help': "Display version"}),
+    (("-s","--show-all-help"),
```
jasonish, line 72: I think I'd rather have this as "--show-advanced" with a help of "Show advanced options".

jasonish (changes requested):
Looking good, I'm liking this cleaned up output.  I think we can hide more behind this advanced flag:
```
--user-agent
--no-check-certificate
--reload-command
--test-command
--url
--local
--etopen
```

TharushiJay:
> Looking good, I'm liking this cleaned up output. I think we can hide more behind this advanced flag:
> 
> ```
> --user-agent
> --no-check-certificate
> --reload-command
> --test-command
> --url
> --local
> --etopen
> ```

Great! I've added these options as well.

suricata/update/parsers.py
```diff
@@ -31,6 +31,11 @@
 
 default_update_yaml = config.DEFAULT_UPDATE_YAML_PATH
 
+show_advanced = False
+
+if "-s" in sys.argv or "--show-all-options" in sys.argv:
```
jasonish, line 36: This will need to be update to --show-advanced.

jasonish (changes requested):
Can you also split the doc change into its own commit? Its a general pattern we follow here. Thanks.

jasonish:
Commit message nit: I think first line should have a little more detail:
```
help ouput: hide advanced options behind a flag
```
or something like that.

TharushiJay:
> Commit message nit: I think first line should have a little more detail:
> 
> ```
> help ouput: hide advanced options behind a flag
> ```
> 
> or something like that.

Sure, done as requested.

suricata/update/parsers.py
```diff
@@ -31,6 +31,11 @@
 
 default_update_yaml = config.DEFAULT_UPDATE_YAML_PATH
 
+show_advanced = False
+
+if "-s" in sys.argv or "--show-advanced" in sys.argv:
+    show_advanced = True
```
jasonish, line 37: Sorry, should have caught this earlier. Not to keen on seeing this in the global module space. I think it fits better as the first thing in the `def parser_arg()`...

jasonish (changes requested):
Oh, and there is still a one line change in the docs in the first code commit.

Thanks!

suricata/update/parsers.py
TharushiJay: I have a small question there. Wouldn't it be necessary to set `show_advanced = True` before `global_arg` and `update_arg` are initialized? When I move it to `def parse_arg()` value for help in the options in `global_arg` are initialized with `argparse_SUPPRESS`. As a result the options are hidden when running `--show-advanced` as well.

Please correct me if I'm wrong.

jasonish: You are correct, I didn't consider they are setup in the module name space.  Please revert.

Also, can you create a new PR for this change per the guidelines instead of a force push please?  Thanks!

TharushiJay:
Replaced with #263

suricata/update/parsers.py
TharushiJay: Replaced with #263

PR #266, opened 2021-01-23
Database

CosmoRied:
-- thank you.

  contribution agreement at
  changes made (if applicable)

Describe changes:
- Implemented a db.py file in suricata-update to allow the updated ruleset to be pushed into a mysql database.
- User should supply connection settings in  a my.cnf file and place it in /etc/suricata to be read by the suricata-update program
- If running suricata-update, the user can optionally supply a --database or --mysqlconf or --database-custom command to have these options run. (--database is enough to create a table in the supplied database and populate it with the rules fetched by suricata-update).
- Database needs to exist in order for the command to run, but anyone can run "create database suricata".
- Fields in database include the raw rule, protocol, header, sid, msg, enabled, priority, gid, classtype, rev.
- Database gets populated with the most recent rules fetched by suricata-update. 
- In future, hopefully suricata can connect to the database and read those rules fetched by suricata-update!

### DIFF
--- a/suricata/update/config.py
+++ b/suricata/update/config.py
@@ -77,6 +77,7 @@ DEFAULT_CONFIG = {
     "enable-conf": "/etc/suricata/enable.conf",
     "drop-conf": "/etc/suricata/drop.conf",
     "modify-conf": "/etc/suricata/modify.conf",
+    "mysql-conf": "/etc/suricata/my.cnf",
     "sources": [],
     LOCAL_CONF_KEY: [],
 
--- a/suricata/update/configs/my.cnf
+++ b/suricata/update/configs/my.cnf
@@ -5,6 +5,7 @@
 
 [client]
 database = suricata_rules
-user = website
-password = W3bPa55
+user = user
+password = password
+host = localhost
 default-character-set = utf8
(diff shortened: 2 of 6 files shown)


CosmoRied:
I need to supply mysql-connector-python package to requirements.txt to continue. Awaiting code review from Jason.

jasonish:
I think this idea is bringing back the idea of plugins for Suricata-Update. The problem with built-in support for a database is that its impossible to meet the needs of a majority of users that might use a database.  This comes down to database choice, for example a lot of users would want PostgreSQL, but then the other half of users would only use MySQL.  Additionally the schema, needs vary so much by user that this is hard to do in a generic fashion as well.

If we had plugin hooks at various, points, in this case an "output plugin", something like this could be a template database plugin and adapted as needed for maximum versatility.

So for now I'm not going to consider database support built directly in, but instead will be looking at ways to provide integration points where it would be easy to extend Suricata-Update (via plugins of sorts) to integrate with a database

CosmoRied:
> I think this idea is bringing back the idea of plugins for Suricata-Update. 

I wasn't here for that, can you please link an example?

> Additionally the schema, needs vary so much by user that this is hard to do in a generic fashion as well.

Not really. Your rule class is pretty good at outlining a case for the schema. 

> but instead will be looking at ways to provide integration points where it would be easy to extend Suricata-Update (via plugins of sorts) to integrate with a database.

Great idea. I look forward to this, as its a much needed aspect of the system's design that I can see from my angle.

PR #286, opened 2021-12-27
Feat 4579/v1:  add fail option and handle it 

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/4579

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -99,6 +99,11 @@ INDEX_EXPIRATION_TIME = 60 * 60 * 24 * 14
 # Rule keywords that come with files
 file_kw = ["filemd5", "filesha1", "filesha256", "dataset"]
 
+def emit(msg):
+    logger.error(msg)
+    if config.args().fail:
+        sys.exit(1)
+
 class Fetch:
 
     def __init__(self):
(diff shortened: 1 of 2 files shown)


suricata/update/main.py
```diff
@@ -1279,8 +1284,8 @@ def _main():
     files = None
 
     if not test_suricata(suricata_path):
-        logger.error("Suricata test failed, aborting.")
-        logger.error("Restoring previous rules.")
+        emit("Suricata test failed, aborting.")
+        emit("Restoring previous rules.")
```
jasonish, line 1288: This should not be a fatal error, even with `--fail` as it will prevent restoring the good rules after a Suricata test failure.

suricata/update/main.py
```diff
@@ -99,6 +99,11 @@
 # Rule keywords that come with files
 file_kw = ["filemd5", "filesha1", "filesha256", "dataset"]
 
+def emit(msg):
+    logger.error(msg)
+    if config.args().fail:
+        sys.exit(1)
+
```
jasonish, line 106: I think a better name for this is `strict_error` or something to signify that it logs an error and optionally exits.

suricata/update/main.py
```diff
@@ -1046,7 +1051,7 @@ def _main():
             logger.info("Found Suricata version %s at %s." % (
                 str(suricata_version.full), suricata_path))
         else:
-            logger.error("Failed to get Suricata version.")
+            emit("Failed to get Suricata version.")
```
jasonish, line 1054: I don't think this should be a fatal error either.

jasonish (changes requested):

jasonish:
I think the scope of what actually fails needs to be tightened up a bit, more in line with whats in https://redmine.openinfosecfoundation.org/issues/4579.

Maybe focus on download errors for now. 

I think there are some cases in this PR that could leave the user in a partially updated ruleset state.

PR #287, opened 2021-12-29
main: use sysconfdir for conf file paths

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/4374

### DIFF
--- a/suricata/update/config.py
+++ b/suricata/update/config.py
@@ -17,6 +17,7 @@
 
 import os.path
 import logging
+from collections import defaultdict
 
 import yaml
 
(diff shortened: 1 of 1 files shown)


jasonish (changes requested):
I don't think we should fall back to `/etc/suricata`.  For example, one could have a Suricata installed from a system package (using /etc/suricata) and another install in `/opt/suricata`.  If the installation in `/opt/suricata` does not have a drop.conf, it should not fall back to the drop.conf found in /etc/suricata (if it exists).

inashivb:
> I don't think we should fall back to `/etc/suricata`. For example, one could have a Suricata installed from a system package (using /etc/suricata) and another install in `/opt/suricata`. If the installation in `/opt/suricata` does not have a drop.conf, it should not fall back to the drop.conf found in /etc/suricata (if it exists).

I see. Should we just exit with an error in that case?

jasonish:
Now included in:
- https://github.com/OISF/suricata-update/pull/309
- https://github.com/OISF/suricata-update/pull/310

PR #295, opened 2022-01-24
handle dataset files properly

inashivb:
- Cover edge cases for invalid dataset rules
- Handle "state" attribute

https://redmine.openinfosecfoundation.org/issues/5010

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -424,14 +424,25 @@ def manage_classification(suriconf, files):
 def handle_dataset_files(rule, dep_files):
     if not rule.enabled:
         return
-    load_attr = [el.strip() for el in rule.dataset.split(",") if "load" in el][0]
-    dataset_fname = os.path.basename(load_attr.split(" ")[1])
+    load_attr = [el.strip() for el in rule.dataset.split(",") if "load" in el]
+    state_attr = [el.strip() for el in rule.dataset.split(",") if "state" in el]
+    if load_attr and state_attr:
+        logger.error("Invalid dataset rule")
+        return
+    elif not load_attr and not state_attr:
+        return
+    elif load_attr:
+        ds_attr = load_attr[0]
+    elif state_attr:
+        ds_attr = state_attr[0]
+
+    dataset_fname = os.path.basename(ds_attr.split(" ")[1])
     filename = [fname for fname, content in dep_files.items() if fname == dataset_fname]
     if filename:
         logger.debug("Copying dataset file %s to output directory" % dataset_fname)
         with open(os.path.join(config.get_output_dir(), dataset_fname), "w+") as fp:
             fp.write(dep_files[dataset_fname].decode("utf-8"))
-    else:
+    elif load_attr:
         logger.error("Dataset file %s was not found" % dataset_fname)
 
 def handle_filehash_files(rule, dep_files, fhash):


suricata/update/main.py
```diff
@@ -424,14 +424,25 @@ def manage_classification(suriconf, files):
 def handle_dataset_files(rule, dep_files):
     if not rule.enabled:
         return
-    load_attr = [el.strip() for el in rule.dataset.split(",") if "load" in el][0]
-    dataset_fname = os.path.basename(load_attr.split(" ")[1])
+    load_attr = [el.strip() for el in rule.dataset.split(",") if "load" in el]
+    state_attr = [el.strip() for el in rule.dataset.split(",") if "state" in el]
```
victorjulien, line 428: "save" is another option, should it be handled here?

jasonish: No, this is about extracting a dataset file included in a rule set.

suricata/update/main.py
```diff
@@ -424,14 +424,25 @@ def manage_classification(suriconf, files):
...
+    if load_attr and state_attr:
+        logger.error("Invalid dataset rule")
+        return
```
jasonish, line 431: Oh, but this might error on a `save` which it shouldn't do.  @inashivb We should probably only log an error if its a dataset rule that needs action, like a load.. Otherwise return early.  We don't need to validate all dataset types here.

suricata/update/main.py
```diff
@@ -424,14 +424,25 @@ def manage_classification(suriconf, files):
...
+        ds_attr = load_attr[0]
+    elif state_attr:
+        ds_attr = state_attr[0]
+
+    dataset_fname = os.path.basename(ds_attr.split(" ")[1])
     filename = [fname for fname, content in dep_files.items() if fname == dataset_fname]
     if filename:
         logger.debug("Copying dataset file %s to output directory" % dataset_fname)
         with open(os.path.join(config.get_output_dir(), dataset_fname), "w+") as fp:
             fp.write(dep_files[dataset_fname].decode("utf-8"))
```
jasonish, line 444: Something we should address with this PR as well is handle the case where the file doesn't exist in the dataset.

suricata/update/main.py
inashivb: Indeed did not think about `save`. Should add tests for all options to avoid these issues.

suricata/update/main.py
inashivb: What is the right action in that case?
Error out and disable the rule?

suricata/update/main.py
inashivb: Probably not since
- the file mentioned with `save` does not already exist in a freshly fetched ruleset
- as a result, nothing needs to be moved along with the rules

suricata/update/main.py
jasonish: For now, just a log an error.

suricata/update/main.py
inashivb: I just realized the checks here are probably not very strict, what is someone has `load,save.csv`, we'll return early since `save` was found in the string. Making it better..

suricata/update/main.py
inashivb: OK I was just looking at this and I need clarification about the case you are mentioning. We are already checking if the dir contains the filename mentioned with the keyword and only then we try to open it.. Do you mean a case like `load` without any filename given to it?

PR #297, opened 2022-03-24
workflows: remove CentOS 8 job as it's EOL

inashivb:


jasonish:
Thanks.  I actually converted it to AlmaLinux 8, its a good replacement for CentOS 8.  Also updated for 1.1.x and 1.2.x branch.

PR #299, opened 2022-03-24
add fail option and handle it - v2

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/4579

Previous PR: https://github.com/OISF/suricata-update/pull/286

Changes since v1:
- Rename `emit` fn to `strict_error`
- Restrict it's scope to failures during fetch only

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -99,6 +99,11 @@ INDEX_EXPIRATION_TIME = 60 * 60 * 24 * 14
 # Rule keywords that come with files
 file_kw = ["filemd5", "filesha1", "filesha256", "dataset"]
 
+def strict_error(msg):
+    logger.error(msg)
+    if config.args().fail:
+        sys.exit(1)
+
 class Fetch:
 
     def __init__(self):
@@ -193,7 +198,7 @@ class Fetch:
             tmp_fileobj.close()
         except URLError as err:
             if os.path.exists(tmp_filename):
-                logger.warning(
+                strict_error(
                     "Failed to fetch %s, "
                     "will use latest cached version: %s", url, err)
                 return self.extract_files(tmp_filename)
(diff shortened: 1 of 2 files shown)


suricata/update/main.py
```diff
@@ -193,14 +198,14 @@ def fetch(self, url):
             tmp_fileobj.close()
         except URLError as err:
             if os.path.exists(tmp_filename):
-                logger.warning(
+                strict_error(
```
inashivb, line 201: This will log an `error` now instead of a warning. I'm thinking it does not go with the message. The message indicates a warning. Should I keep it a warning or modify the message a bit?

jasonish: Thats fine. Its probably better as an error than a warning anyways.

suricata/update/main.py
```diff
@@ -193,14 +198,14 @@ def fetch(self, url):
...
                     "Failed to fetch %s, "
                     "will use latest cached version: %s", url, err)
```
jasonish, line 203: I think we need a conditional here and log differently.  If not strict, the message is valid as we'll use the latest cached version.  If strict we're not going to use the cached version as its a fatal error.

suricata/update/main.py
```diff
@@ -216,7 +221,7 @@ def run(self, url=None):
                 files.update(fetched)
             except URLError as err:
                 url = url[0] if isinstance(url, tuple) else url
-                logger.error("Failed to fetch %s: %s", url, err)
+                logger.error("Failed to fetch {}: {}".format(url, err))
```
jasonish, line 224: This is the error line that is hit if we get a 404.. So a strict check needs to be done here.

jasonish (changes requested):
My test command does not fail:
```
$ ./bin/suricata-update --fail -D ./data --suricata /opt/suricata/6.0.5-dev/bin/suricata --no-test --url https://rules.emergingthreats.net/open/suricata-6.0.5/emerging.rules.tar.gz --url http://codemonkey.net/does/not/exist
```

It logs the following error, but still continues:
```
28/3/2022 -- 17:09:42 - <Error> -- Failed to fetch http://codemonkey.net/does/not/exist: HTTP Error 404: Not Found
```

PR #301, opened 2022-03-24
Allow user to log detailed rule report

inashivb:
Add a `--report <filename>` option so that the end user is able to log a
report about the rules enabled, disabled, modified and dropped.
Usage
```
$ suricata-update --report /tmp/rules.log
```

Sample Report

```
Suricata-Update Summary - Thursday, 24 Mar 2022, 13:20:30

Summary
=======
Rules disabled by disable.conf: 148
Rules enabled by enable.conf: 0
Rules modified by modify.conf: 0
Rules converted to drop: 0
Rules enabled for flowbit dependencies: 131
Rules added: 274
Rules removed: 17
Rules modified: 1721

Added Rules
-----------
[1:2500036] ET COMPROMISED Known Compromised or Hostile Host Traffic group 19 (compromised.rules)
[1:2500038] ET COMPROMISED Known Compromised or Hostile Host Traffic group 20 (compromised.rules)

```
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2256

Previous PR: https://github.com/OISF/suricata-update/pull/300

Changes since v9:
- Updated changed variable name all through the function

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -33,6 +33,7 @@ import tempfile
 import signal
 import errno
 from collections import namedtuple
+from datetime import datetime as dt
 
 try:
     # Python 3.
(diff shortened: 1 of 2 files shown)


gordnhoo:
Hello don't know if this already exists but why not also adding to the logging if the update happened successfully or not ?I think this would be a nice addition

jasonish:
Success is reported by the exit status, so there are a few way to achieve this.  If using `cron`, `cron` will typically log the failure and output. If already processing your system logs, it should be searchable.

If not using `cron`, but some other tool to schedule the update, that should then check the exit status and report error/success.

Alternatively, the output could be redirected to a file, then the log checked for "Warning" or "Error" messages.

Ultimately the best solution probably depends on how you wish to be notified of errors. But I believe Suricata-Update provides enough for the reporting of the errors already.

inashivb:
Closing as stale.

PR #302, opened 2022-03-24
use cookie to ensure write to a valid rule dir

inashivb:
Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket: https://redmine.openinfosecfoundation.org/issues/2691

Previous PR: https://github.com/OISF/suricata-update/pull/157

Changes since v1:
Entirely revamped. Used cookie approach suggested by Jason on Redmine

Known TODOs:
- Same needs to be done for merged rule file

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -509,6 +509,13 @@ def write_to_directory(directory, files, rulemap, dep_files):
     modified = []
 
     oldset = {}
+
+    cookie_path = os.path.join(directory, ".suricata-update")
+    if os.path.exists(directory):
+        if not os.path.exists(cookie_path) and os.listdir(directory):
+            logger.error("{} -- doesn't look like a suricata-update rule directory".format(directory))
+            sys.exit(1)
+
     if not args.quiet:
         for file in files:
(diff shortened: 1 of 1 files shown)


suricata/update/main.py
```diff
@@ -509,6 +509,13 @@ def write_to_directory(directory, files, rulemap, dep_files):
     modified = []
 
     oldset = {}
+
+    cookie_path = os.path.join(directory, ".suricata-update")
+    if os.path.exists(directory):
+        if not os.path.exists(cookie_path) and os.listdir(directory):
+            logger.error("{} -- doesn't look like a suricata-update rule directory".format(directory))
+            sys.exit(1)
```
inashivb, line 517: @jasonish do you think this is as per your expectations?
Questions:
1. How do we ensure that this does not break the current setups? Because we don't have a cookie already in the rule dir.
2. Is checking for dir good enough for determining  "first run"?
3. Should the cookie be in say `<prefix>/suricata` or `<prefix>/suricata/rules`

inashivb:
Closing as stale.

PR #298, opened 2022-03-24
handle dataset files properly - v3

inashivb:
- Cover edge cases for invalid dataset rules
- Handle "state" attribute
- Make checks more robust

Link to [redmine](https://redmine.openinfosecfoundation.org/projects/suricata/issues) ticket:
https://redmine.openinfosecfoundation.org/issues/5010

Previous PR: https://github.com/OISF/suricata-update/pull/295

Changes since v2:
- Proper handling of `save` and `state`

PR #303, opened 2022-03-29
dataset handling - v4

jasonish:
A continuation of https://github.com/OISF/suricata-update/pull/298.

Ticket: https://redmine.openinfosecfoundation.org/issues/5010

Changes from last PR:
- Fixes path handling.
- Saves the dataset file as the md5 hash of the contents and rewrites the rule accordingly to prevent duplicate dataset filenames among rulesets overwriting each other.

I'm not wondering if it would be more use friendly to put the dataset files in a directory named after the source, instead of using a hash as the filename. That would cover the duplicate case better.

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -424,26 +424,32 @@ def manage_classification(suriconf, files):
 def handle_dataset_files(rule, dep_files):
     if not rule.enabled:
         return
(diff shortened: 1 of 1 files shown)


jasonish:
Now included in:
- https://github.com/OISF/suricata-update/pull/310
- https://github.com/OISF/suricata-update/pull/309

PR #304, opened 2022-03-29
Modify filter to add new metatdata - v1

jasonish:
A new addition to `modify.conf` that allows for adding a new metadata.
    
Example configuration lines:
      metadata-add re:"SURICATA STREAM" "evebox.action" "archive"
      metadata-add 2010646 "evebox.action" "archive"
    
Matching rules will have a new `metadata` section added onto the end of the rule.

Ticket: https://redmine.openinfosecfoundation.org/issues/5221


jasonish:
> I've tested `metadata-add` with evebox to auto archive things and it works well.
> 
> Nit: `add-metadata` in commit msg, `metadata-add` as the actual option.

Thanks. Addressed in https://github.com/OISF/suricata-update/pull/306

PR #318, opened 2022-09-01
1.2.x update entrypoint/v1

jasonish:
- entry point: update sys.path for non-distutils install
- changelog: update


victorjulien:
Merged after rebase, thanks!

PR #320, opened 2022-11-06
Match on rule metadata - v1

jasonish:
Allow metadata matching for enable and disable. For example:

    metadata: deployment perimeter

will match rules with "metadata: deployment Perimeter".

Match is case insensitive.

Ticket: https://redmine.openinfosecfoundation.org/issues/5561


jasonish:
Merged.

PR #321, opened 2022-11-23
Allow checksum URL to be provided in the index - v1

jasonish:
- update: allow index "checksum" value to be a url
- github-ci: fix pytest on MacOS


jasonish:
Merged.

PR #330, opened 2023-07-05
conf: use correct configuration files - v1

jasonish:
Ticket: https://redmine.openinfosecfoundation.org/issues/6172

If /etc/suricata/update.yaml specific a "disable-conf" of
"/some/path/to/dummy.conf", but /etc/suricata/disable.conf existed, it prefered
the existence of the file over the user specific one. This did not affect
command line parameters.

Now when looking for the default files, only override the configuration file
values if they are not present in the configuration file. This retains the
default behaviour of auto-finding the files, but respects the configuration
over auto-finding.

This also allows us to display an error if a specified configuration file does
not exist.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -166,30 +166,6 @@ jobs:
       - name: Python 3 integration tests
         run: PYTHONPATH=. python3 ./tests/integration_tests.py
 
-  ubuntu-1604:
-    name: Ubuntu 16.04
-    runs-on: ubuntu-latest
-    container: ubuntu:16.04
-    steps:
-      - run: apt update
-      - run: |
-          apt -y install \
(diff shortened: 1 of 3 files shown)


inashivb (approved):
Tests OK for me. 👍🏽 

Q: Should we also add in logs which `update.yaml` is being used in the current run if it was not given with `-c` on cmdline?

jasonish:
> Tests OK for me. 👍🏽
> 
> Q: Should we also add in logs which `update.yaml` is being used in the current run if it was not given with `-c` on cmdline?

Which `update.yaml` should be logged as one of the first things, provided one is found:
```
6/7/2023 -- 17:16:36 - <Info> -- Loading /etc/suricata/update.yaml
```

PR #37, opened 2018-06-02
Support for multiple conditions to match a rule

securitym0nkey:
Make sure these boxes are signed before submitting your Pull Request
-- thank you.

- [x] I have read the contributing guide lines at
  https://redmine.openinfosecfoundation.org/projects/suricata/wiki/Contributing
- [xI have signed the Open Information Security Foundation
  contribution agreement at
  https://suricata-ids.org/about/contribution-agreement/
- [x] I have updated the user guide (in doc/userguide/) to reflect the
  changes made (if applicable)

[redmine issue 2509](https://redmine.openinfosecfoundation.org/issues/2509):

Describe changes:

This adds support to specify multiple conditions to match a rule in
{enable,disable}.conf.

Syntax:
multi:<match_condition1>;<match_conditios2>;...<match_conditionN>;

Example:
Enable all rules including the term "nmap" bust just from the "emerging-scan.rules" file.

multi:filename:rule/emerging-scan.rules; re:nmap;

PR #331, opened 2023-07-31
matching: multi match support

securitym0nkey:
-- thank you.

  contribution agreement at
  changes made (if applicable)

This is an implementation of  [redmine 2509](https://redmine.openinfosecfoundation.org/issues/2509)

Describe changes:

This is a resubmit of pr #37. After reworking to fit the 2023 code base.

It adds support to specify multiple conditions to match a rule.

Syntax:
`multi:<match_condition1>;<match_condition2>;...<match_conditionN>;`

Examples:
Match all rules including the term "nmap" but just from the "emerging-scan.rules" file.
`multi:filename:rule/emerging-scan.rules; re:nmap;`

Match all rules with a recent cve reference and a perimeter deployment
`multi:re:cve-202[23];metadata: deployment perimeter;`

### DIFF
--- a/doc/update.rst
+++ b/doc/update.rst
@@ -259,6 +259,15 @@ with the provided metadata::
           convert rules to drop. It is not available for rule
           modification.
 
+Multi Matching
+--------------
+
+The above matching methods can also be combined. It's logical AND, so all combined matcher need to match a given rule::
+
+  multi:filename:*/emerging-scan.rule;re:nmap;
+  multi:group:emerging-web_specific_apps;re:wordpress;re:cve[-,]201[6-9];
+  multi:re:cve[-_]202[23];metadata: deployment perimeter;
+
 Modifying Rules
 ---------------
 
--- a/suricata/update/configs/disable.conf
+++ b/suricata/update/configs/disable.conf
@@ -17,3 +17,7 @@
 # Disable all rules with a metadata of "deployment perimeter". Note that metadata
 # matches are case insensitive.
 # metadata: deployment perimeter
+
+# Example of multi matching
+# Disable all rules with significant performance impact (metadata match) from emerging-policy (group match)
+multi:metadata: performance_impact Significant;group:emerging-policy;
--- a/suricata/update/configs/drop.conf
+++ b/suricata/update/configs/drop.conf
@@ -9,3 +9,7 @@
 #
 # re:heartbleed
 # re:MS(0[7-9]|10)-\d+
+
+# Example of multi matching
+# Set phishing related (metadata match) rules from emerging-current_events (group match) to drop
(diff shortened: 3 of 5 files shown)


jasonish:
Sorry for taking so long to get to this.  One issue I have is not all valid regex's are supported.  For example:

```
re: sid:\s*2100498;\s*rev:7;
```

works...

but:

```
multi: re: sid:\s*2100498;\s*rev:7;
```

appears to work, but actually fails to properly load as it will only take up to the `;`, then silently fail to load the remainder. So I think that would need to be addressed.

jasonish (changes requested):
https://github.com/OISF/suricata-update/pull/331#issuecomment-2580917207

securitym0nkey:
That's a good point. `;` is definitely not a good delimiter. So what could be a better one? Almost every character might be part of a valid regex, right? Though other delimiters might be less likely to be used in someones regex. 

My quick idea would be to use a **[TAB]** as a delimiter. A good regex would use `\t` instead of a real **[TAB]**. But i'm not really satisfied with that solution myself. Any thoughts on what would be acceptable. 

Certainly its a good idea to throw some warning or error if a child matcher is not a valid one. I think that would happen with with your example.

jasonish:
> That's a good point. `;` is definitely not a good delimiter. So what could be a better one? Almost every character might be part of a valid regex, right? Though other delimiters might be less likely to be used in someones regex.
> 
> My quick idea would be to use a **[TAB]** as a delimiter. A good regex would use `\t` instead of a real **[TAB]**. But i'm not really satisfied with that solution myself. Any thoughts on what would be acceptable.
> 
> Certainly its a good idea to throw some warning or error if a child matcher is not a valid one. I think that would happen with with your example.

Yeah, I'm not sure either. It was not created with more parameters on a line than one in mind... Proper quoting could work, but also a breaking change as we don't currently require quoting now.

securitym0nkey:
After some thoughts i would suggest to use ` && ` **[space][andsign][andsign][space]** as the delimiter. I can't come up with a situation where this causes a conflict with other matchers except when the regular expression as exactly ` && ` in it. I can't think of a situation where one would write such regex. Obviously it needs to be documented as invalid when using the multi matcher. The workaround would be to encode ` && ` as ` \x26\x26 ` within the regex.

So your example combined with some other matchers would look something like this.
```
multi: re: sid:\s*2100498;\s*rev:7; && multi:filename:rule/emerging-scan.rules && metadata: deployment perimeter
```

Why i like this solution:
- no quoting, no breaking change
- && indicates that the multiple matchers are logical AND
- even the edge case where the delimiter is part of the regex can be addressed

If that's an acceptable solution i would go ahead and implement it.

PR #335, opened 2023-11-16
update min-version required

inashivb:
Jason, who are the consumers of `min-version` keys? https://github.com/OISF/suricata-intel-index/blob/master/index.yaml#L59
Do we always know that these external sources do support the said min-versions even now?

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -88,7 +88,7 @@ else:
     logger = logging.getLogger()
 
 # If Suricata is not found, default to this version.
-DEFAULT_SURICATA_VERSION = "4.0.0"
+DEFAULT_SURICATA_VERSION = "6.0.0"
 
 # The default filename to use for the output rule file. This is a
 # single file concatenating all input rule files together.
--- a/doc/quickstart.rst
+++ b/doc/quickstart.rst
@@ -120,7 +120,7 @@ This command will:
   and do not need to exist.
 
 * Download the Emerging Threats Open ruleset for your version of
-  Suricata, defaulting to 4.0.0 if not found.
+  Suricata, defaulting to 6.0.0 if not found.
 
 * Apply enable, disable, drop and modify filters as loaded above.
 


suricata/update/main.py
```diff
@@ -88,7 +88,7 @@
     logger = logging.getLogger()
 
 # If Suricata is not found, default to this version.
-DEFAULT_SURICATA_VERSION = "4.0.0"
```
jufajardini, line 91: Should this deserve an upgrading note somewhere, considering that ET, for instance, has rulesets for older versions, so this might affect some of their folks? 🤔

jasonish:
> Jason, who are the consumers of `min-version` keys? https://github.com/OISF/suricata-intel-index/blob/master/index.yaml#L59

Suricata-Update, and if its not looking at it, that was the idea.  So if a ruleset had a `min-version` of 7, but you were running 6 it wouldn't be presented as an option to enable.

jasonish:
> Do we always know that these external sources do support the said min-versions even now?

We have no idea.

suricata/update/main.py
jasonish: I think this is fine, we don't have an upgrade section and its easy to work-around this.

jasonish (approved):

victorjulien (approved):

jasonish:
Found this ticket https://redmine.openinfosecfoundation.org/issues/4930, I wonder if we should do that at some point as well?

inashivb:
> Found this ticket https://redmine.openinfosecfoundation.org/issues/4930, I wonder if we should do that at some point as well?

Indeed.
Q: Would you rather see this ticket done than this patch?
Request: Could you please comment on the ticket what should be the default version? For example,  in https://github.com/OISF/suricata-intel-index/blob/master/index.yaml#L232, `recommended` or `6.0`?

victorjulien:
Are there risks to existing users?

victorjulien:
> > Are there risks to existing users?
> 
> Only to users who are running Suricata-Update unbundled from Suricata, on a system with no Suricata in their PATH and are expecting to get 4.0.0 rules. So potentially.

Seems low enough risk to me, so lets merge this.

jasonish:
Merged via https://github.com/OISF/suricata-update/pull/339. Thanks!

PR #340, opened 2024-02-29
fetch: handle bare file urls not ending in .rules

jasonish:
If a URL is a bare file, and does not end in .rules, Suricata-Update will ignore it.  Such URLs might look like:

- https://<misp>/1011

Tested with https://codemonkey.net/tmp/1011

Ticket: https://redmine.openinfosecfoundation.org/issues/3664


inashivb (approved):
Tests well. LGTM 🚀

jasonish:
Merged.

PR #348, opened 2024-11-27
matching: use a set match for SIDs instead of an array - v1

jasonish:
- **matching: consolidate sid matchers into a set matcher**
  Consolidate SID matchers into a single SID set matcher which stores a
  dict of all SIDs to be matched. An array of many SID matchers to a
  single matcher with much faster lookup.
  
  This can reduce a many minute runtime down to 10s of seconds.
  
  Ticket: https://redmine.openinfosecfoundation.org/issues/7415
  

- **github-ci: remove centos-7 build, EOL**
  

- **github-ci: add fedora 40, 41; remove 38, 39**
  

- **github-ci: run macos python jobs in virtualenv**
  With the latest brew changes, a virtualenv is required to install
  pyyaml.

### DIFF
--- a/suricata/update/matchers.py
+++ b/suricata/update/matchers.py
@@ -51,6 +51,25 @@ class ProtoRuleMatcher:
         return rule.proto == self.proto
 
 
+class IdSetRuleMatcher(object):
+    """Matcher object that matches on a set of rule SIDs.
+
+    This matcher is not directly parsed, but instead constructed after
+    loading individual IdRuleMatchers and consolidated into one
+    matcher that is much faster.
+
+    """
+    def __init__(self):
+        self.sids = {}
+
+    def add(self, generatorId, signatureId):
+        self.sids[(generatorId, signatureId)] = True
+
+    def match(self, rule):
+        key = (rule.gid, rule.sid)
+        return key in self.sids
+
+
 class IdRuleMatcher(object):
     """Matcher object to match an idstools rule object by its signature
     ID."""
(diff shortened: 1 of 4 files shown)


jasonish:
The MacOS CI failure is about the pyyaml module, not this PR.

suricata/update/matchers.py
```diff
@@ -51,6 +51,25 @@ def match(self, rule):
         return rule.proto == self.proto
 
 
+class IdSetRuleMatcher(object):
+    """Matcher object that matches on a set of rule SIDs.
+
+    This matcher is not directly parsed, but instead constructed after
+    loading individual IdRuleMatchers and consolidated into one
+    matcher that is much faster.
```
inashivb, line 59: Q: Why don't we parse directly into this Matcher?

inashivb (commented):
Looks good.
1. Question inline.
2. Shall we get rid of the MacOS commit altogether if it's not fixing anything anyway?

inashivb (approved):
Let's merge this! 🚀

suricata/update/matchers.py
```diff
@@ -51,6 +51,25 @@ def match(self, rule):
...
+
+    """
+    def __init__(self):
+        self.sids = {}
+
+    def add(self, generatorId, signatureId):
+        self.sids[(generatorId, signatureId)] = True
```
victorjulien, line 66: side note, I think we should allow specifying a rev here too. Recently I had a case where a ruleset pushed a broken rule, so I wanted to just disable the broken rev

victorjulien (approved):

PR #354, opened 2025-04-08
fixups/v2

jasonish:
- **tests: add tox tests for python 3.12 and 3.13**
  

- **cleanups: apply valid cleanups suggested by copilot**
  Add test for gid/sed/rev that shouldn't match.
  

- **md5: don't rely on version for usedforsecurity**
  The usedforsecurity flag to hashlib.md5 doesn't appear to be version
  specific, due to backporting, etc.
  
  Instead attempt to use it, on exception, fallback to not using it.
  
  Ticket: https://redmine.openinfosecfoundation.org/issues/7255
  

- **config: handle suricata config with '=' in value**
  For example, when an '=' appears in the --dump-config output when its
  used as part of a bpf-filter.
  
  Ticket: https://redmine.openinfosecfoundation.org/issues/7637

### DIFF
--- a/suricata/update/engine.py
+++ b/suricata/update/engine.py
@@ -109,7 +109,7 @@ class Configuration:
         conf = {}
         for line in configuration_dump.splitlines():
             try:
-                key, val = line.decode().split(" = ")
+                key, val = line.decode().split(" = ", maxsplit=1)
                 conf[key] = val
             except:
                 logger.warning("Failed to parse: %s", line)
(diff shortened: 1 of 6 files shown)


victorjulien:
Is the failing MacOS check something to be concerned about?

inashivb (approved):
LGTM 🚀

jasonish:
> Is the failing MacOS check something to be concerned about?

The code should run on Mac, provided the Python dependencies can be installed properly. That is what is wrong on the Mac build, I can't properly get the dependencies installed, and have been unable to resolve it via CI.

victorjulien:
Should we remove the check then? Leaving it in a broken state will increase risk of overlooking real CI issues, as we come to expect to see, then ignore failure.

PR #357, opened 2025-07-02
rules: fix parsing of address lists

jasonish:
The previous parser just looked for the next "]" to find the end of a
list without respect for list depth. Instead step through the array
tracking the depth of the nested lists.

Ticket: #7799


jasonish:
Merged with https://github.com/OISF/suricata-update/pull/358.

PR #365, opened 2025-10-01
20251001/staging/v1

jasonish:
- **matchers: fix regular expression matching**
  A regular express with multiple ':' was accidentally being parsed as an
  ID matcher. Making ID matching more strict.
  
  Ticket: https://redmine.openinfosecfoundation.org/issues/7922
  

- **misc: count number of rules that were actually modified**
  And not all that are just subject to modification.
  
  Ticket: https://redmine.openinfosecfoundation.org/issues/7967
  

- **tox: update versions for integration tests**

### DIFF
--- a/suricata/update/main.py
+++ b/suricata/update/main.py
@@ -1304,8 +1304,10 @@ def _main():
 
         for fltr in modify_filters:
             if fltr.match(rule):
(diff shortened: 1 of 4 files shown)


victorjulien (approved):
Tested OK

